PR #1, opened 2015-09-10
The Distance Method HGT detection algorithm implemented in python

ekopylova:
Implementation of the Distance Method for HGT detection based on algorithm described in:
        Wei. X et al., "A Distance-Based Method for Detecting HGT in Whole
        Genomes", International Symposium on Bioinformatics Research and
        Applications (ISBRA), 2008, pages 26-37

```
The workflow follows the algorithm:
1. For each gene in target genome,
    i.    BLAST sequence against all other genes in the reference genomes;
    ii.   Go to step 3 if gene has more than threshold number of homologs
          (min-num-homologs), otherwise go to next gene in target genome;
    iii.  Compute multiple sequence alignment on homolog genes using
          CLUSTAL;
    iv.   Compute pairwise distance matrix using PHYLIP's protdist
          function and Z-score normalize the set of pairwise distances
          for each gene family and species;
    v.    Add distance matrix for all pairwise distances into global
          distance matrix storing results for all genes

2. Cluster gene families by species,
    vi.   Compute all species sets (sets of genes whose orthologs are
          detectable in exactly the same subset of the considered
          species);
    vii.  Cluster genes to each core species set cluster using the Hamming
          distance clustering algorithm;
    viii. Run outlier detection algorithm on each cluster (paragraph 2
          of section 'Detecting Outlier Genes' in original paper)
```


distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+
+# ----------------------------------------------------------------------------
+# Copyright (c) 2015--, The WGS-HGT Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+""" Implement Distance Method for HGT detection based on algorithm described
```
antgonza: These should be comments not a string, right?

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+import shlex
+from os import listdir
+from os.path import join, splitext, basename
+from itertools import imap
+
+from skbio.parse.sequences import parse_fasta
+
+
+class Command(object):
+    """
```
antgonza: Generally we like to have only 1 line as the main description to follow the numpy standard

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+    Enables to run subprocess commands in a different thread
+    with TIMEOUT option.
+
+    Based on jcollado's solution:
+    http://stackoverflow.com/questions/1191374/subprocess-with-timeout/4825933#4825933
+    https://gist.github.com/kirpit/1306188
+    """
+    command = None
```
antgonza: This is generated at init so you shouldn't need it here. Same with the other variables.

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+        if thread.is_alive():
+            self.process.terminate()
+            thread.join()
+        return self.status, self.output, self.error
+
+
+def hamming(str1, str2):
+    "Compute the Hamming distance between two strings"
+    assert len(str1) == len(str2)
+    ne = operator.ne
```
antgonza: I think you can delete this line and then:
`sum(imap(operator.ne, str1, str2))`

distance-method/distance_method.py
antgonza: Note that this applies to all comments.

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+    return sum(imap(ne, str1, str2))
+
+
+def preprocess_data(working_dir,
+                    target_proteomes_dir,
+                    verbose=False):
+    """ Map each gene to sudo name (ex. 1_1 for species 1, gene 1) for easier
```
antgonza: The explanation should go in a notes section. Also, why the limitation?

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+    ref_db: dictionary
+      dictionary storing FASTA label as key and sequence as value for the
+      reference databases
+    species: integer
+      the number of species in the reference databases
+    """
+    gene_map = {}
+    ref_db = {}
+    if verbose:
+        sys.stdout.write("Target organism\tNumber of genes\n")
```
antgonza, line 148: Why use `sys.stdout.write` vs `print`?

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+    # each file contains genes for species
+    species = 0
+    for _file in listdir(target_proteomes_dir):
```
antgonza: This is gonna fail if there are other files in that folder, right? Like .DS_Store ...

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+        if verbose:
+            sys.stdout.write("%s. %s\t" % (species+1, basename(_file)))
+        with open(join(target_proteomes_dir, _file), 'rb') as readfile:
+            gene = 0
+            for label, seq in parse_fasta(readfile):
+                label = label.split()[0]
+                ref_db[label] = seq
+                sudo_label = "%s_%s" % (species, gene)
+                gene_map[label] = sudo_label
```
antgonza: Will it be worth verifying that the name doesn't exist before adding it? Otherwise it will overwrite values silently.

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+    if verbose:
+        sys.stdout.write("Target organism\tNumber of genes\n")
+    # each file contains genes for species
+    species = 0
+    for _file in listdir(target_proteomes_dir):
+        if verbose:
+            sys.stdout.write("%s. %s\t" % (species+1, basename(_file)))
+        with open(join(target_proteomes_dir, _file), 'rb') as readfile:
+            gene = 0
+            for label, seq in parse_fasta(readfile):
```
antgonza: To count gene and species it will be easier to use enumerate.

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+                            stderr=subprocess.PIPE,
+                            close_fds=True)
+    proc.wait()
+    stdout, stderr = proc.communicate()
+    if (stderr and debug):
+        print "[DEBUG] %s\n" % stderr
+
+    # launch blast
+    out_file_fp = join(
+        working_dir, "%s.blast" % basename(splitext(query_proteome_fp)[0]))
```
antgonza: Why not use basename?

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+      names as values, and vica versa
+    debug: boolean
+      if True, run function in debug mode
+    """
+    # read blastp results
+    with open(alignments_fp, 'U') as alignments_f:
+        for line in alignments_f:
+            if debug:
+                sys.stdout.write("[DEBUG] %s" % line)
+            line = line.split()
```
antgonza: Are these always 2 elements? If so you could: `query, ref = line.split()`. Note that if you get more elements this line is gonna break

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+            query = line[0]
+            ref = line[1]
+            if query not in hits:
+                hits[query] = [ref]
+            else:
+                # check that the query mapped to a different species
+                # since we only want the best homolog per species
+                current_species = gene_map[ref].split('_')[0]
```
antgonza: What about changing this block to:

```
if [gene_map[gene].split('_')[0] for gene in hits[query] if gene_map[gene].split('_')[0] == current_species]:
    hits[query].append(ref)
```

or

```
if current_species in [gene_map[gene].split('_')[0] for gene in hits[query]]:
    hits[query].append(ref)
```

or something similar?

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+      dictionary storing query (gene) names as keys and the best aligning
+      reference sequences as values (one alignment per reference sequence)
+    query: string
+      query gene name
+    timeout: integer
+      number of seconds to allow Clustalw to run before terminating the
+      process
+    """
+    with open(fasta_in_fp, 'w') as in_f:
+        sorted(hits[query])
```
antgonza: Why sorted?

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+                "status: %s\noutput: %s\terror: %s\t" % (
+                    status, output, error))
+
+
+def compute_distances(phylip_command_fp,
+                      warnings=False):
+    """ Compute distances between each pair of sequences in MSA (multiple
+        sequence alignment) using PHYLIP's protdist function.
+
+        Parameters
```
antgonza: Lines have extra tab.

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+        dictionary containing the binary indicator vectors as keys and the
+        number of genes with identical species set represented by the binary
+        vectors as values
+    gene_bitvector_map: list
+        list containing the binary indicator vector for each query gene
+    debug: boolean
+        if True, run function in debug mode
+    """
+    # assume a pairwise alignment exists for all species
+    missing_species = [str(x) for x in range(0, num_species)]
```
antgonza, line 424: This can be added to `preprocess_data` and you could avoid looping twice, right?

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+    with open(phylip_fp, 'U') as phylip_f:
+        for line in phylip_f:
+            alignment_dist = line.strip().split()
+            if len(alignment_dist) == 1:
+                continue
+            if not line.startswith(' '):
+                species = line.split()[0].split('_')[0]
+                missing_species.remove(species)
+
+    # scan through file again, collecting alignment
```
antgonza, line 434: Same with this. Note that if you can't not join with `preprocess_data` you could join these 2 sections, right?

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+    orig_order_labels = []
+    p = numpy.empty(shape=(num_species, num_species))
+    p.fill(numpy.nan)
+    idx_p = 0
+    ind_a = 0
+    with open(phylip_fp, 'U') as phylip_f:
+        alignment_list = []
+        # skip first line containing number of lines in
+        # the file
+        next(phylip_f)
```
antgonza, line 444: You are using two different methods to skip the 1st line. Look at 419-420

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+                sys.stdout.write("[DEBUG] %s" % line)
+            alignment_dist = line.strip().split()
+            if line.startswith(' '):
+                alignment_list.extend(alignment_dist)
+            else:
+                # new species alignment pairs
+                if alignment_list:
+                    for i in range(0, len(missing_species)):
+                        alignment_list.append(None)
+                    a = numpy.asarray(alignment_list[1:], dtype=float)
```
antgonza, line 456: Can this block become:
`a = (a - numpy.nanmean(a)) / numpy.nanstd(a)`

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+                    # modify array with normalized values
+                    for x in numpy.nditer(a, op_flags=['readwrite']):
+                        x[...] = (x-mean)/stdev
+                    p[idx_p] = a
+                    idx_p += 1
+                    orig_order_labels.append(alignment_list[0])
+                alignment_list = alignment_dist
+
+    # add distance on final line
+    for i in range(0, len(missing_species)):
```
antgonza, line 464: Same here.

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+    # update species set counts
+    if bitvector_gene not in species_set_dict:
+        species_set_dict[bitvector_gene] = 1
+    else:
+        species_set_dict[bitvector_gene] += 1
+
+    gene_bitvector_map[full_distance_matrix_offset] = bitvector_gene
+
+    # sort the distance matrix based on species names (S1, S2, S3 ..)
+    # in order to be consistent across all gene families
```
antgonza, line 489: This section seems that it could be improved by using argsort

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+    # with more than species_set_size genes)
+    gene_clusters_dict = {}
+    # if the largest species set contains less than
+    # threshold (species_set_size) elements, set the
+    # only core cluster to the largest species set
+    if sorted_species_set[0][1] < species_set_size:
+        gene_clusters_dict[sorted_species_set[0][0]] = []
+    else:
+        for bitvector in sorted_species_set:
+            if bitvector[1] >= species_set_size:
```
antgonza: Is it OK not to have an else?

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+                gene_clusters_dict[bitvector[0]] = []
+
+    # assign species sets with fewer than species_set_size
+    # species to core clusters if the Hamming distance
+    # between the two bitvectors is less than
+    # hamming_distance
+    species_set_assigned = []
+    for cluster_core in gene_clusters_dict:
```
antgonza, line 575: In a lot of this code you could reduce the number of indexing by using temporal variables. Perhaps something like?

```
for cluster_core in gene_clusters_dict:
    bv = bitvector[0]
    if bv not in species_set_assigned and hamming(cluster_core, bv) <= hamming_distance:
         gene_clusters_dict[cluster_core].append(bv)
         species_set_assigned.append(bv)
```

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+            t[map_orig_sorted[idx_b]] = el
+        p2[map_orig_sorted[idx_a]] = t
+    del p
+
+    # add normalized distance matrix for current gene
+    # to full distance matrix
+    full_distance_matrix[full_distance_matrix_offset] = p2
+
+
+def cluster_distances(species_set_dict,
```
antgonza, line 506: I believe this function can be reworked so you do not have to iterate several times on `for bitvector in sorted_species_set:`

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+    distance_vector = numpy.zeros(total_genes)
+    if debug:
+        sys.stdout.write("[DEBUG] species_species\t")
+        for k in range(total_genes):
+            sys.stdout.write("gene # %s".ljust(12) % k)
+        sys.stdout.write("[low_bound, up_bound]\n")
+    for i in range(num_species):
+        for j in range(num_species):
+            if i != j:
+                for k in range(total_genes):
```
antgonza, line 678: I think you can join this loop with the one below `for k, distance in enumerate(distance_vector)`

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+                            sys.stdout.write(
+                                "%s\033[92m%s\033[0m" % (spaces, distance))
+                    elif debug:
+                        sys.stdout.write("%s%s" % (spaces, distance))
+                if debug:
+                    sys.stdout.write("\t[%s, %s]\n" % (low_bound, up_bound))
+
+    # traverse outlier_matrix by gene and count the number of
+    # outlier distances by species
+    outlier_count_matrix = numpy.zeros(
```
antgonza, line 703: You could also join the next two for loops.

distance-method/distance_method.py
```diff
@@ -0,0 +1,981 @@
...
+            if i != j:
+                for k in range(total_genes):
+                    distance_vector[k] = full_distance_matrix[k][i][j]
+                mean = numpy.nanmean(distance_vector)
+                stdev = numpy.nanstd(distance_vector)
+                low_bound = round(mean - stdev_offset*stdev, 5)
+                up_bound = round(mean + stdev_offset*stdev, 5)
+                if debug:
+                    sys.stdout.write("[DEBUG] %s_%s\t".ljust(20) % (i, j))
+                for k, distance in enumerate(distance_vector):
```
antgonza, line 686: My guess is that at the end of this for you could add the next two fors (704-719) and avoid some loops

antgonza:
In general, I think some of the blocks can be improved to avoid so many loops, which should make the code run faster. Suggest making those changes on the main code without changing the tests so you make sure the results are not changing.

distance-method/distance_method.py
ekopylova: The maximum 10 character taxon ID is the [Phylip MSA format](http://www.bioperl.org/wiki/PHYLIP_multiple_alignment_format). The species_genes naming is limited to 9999_99999 to comply with the max 10 char format.

distance-method/distance_method.py
ekopylova, line 148: Mainly to avoid newlines with print (ex. in the preprocess_data() function), maintain consistent syntax throughout the code and because it's more C-like which was my first solution to avoid the newlines (though I see with Python 3, the print statement is a function and I could do the same there)

distance-method/distance_method.py
ekopylova: True. Changed this to process files with default extensions .fa, .fasta and .faa and any extra extensions provided by the user (via new option --ext).

distance-method/distance_method.py
ekopylova: Done.

distance-method/distance_method.py
antgonza: Thanks!

distance-method/distance_method.py
ekopylova: Yes.

distance-method/distance_method.py
ekopylova: Didn't want the original extension in the name, but it does make the code look cleaner. Changed.

distance-method/distance_method.py
ekopylova: Yup, it's a 12-field output and I only need the first two.

distance-method/distance_method.py
ekopylova: Done.

distance-method/distance_method.py
ekopylova: Thanks! The second option looks good, but instead `if current_species not in` ..

distance-method/distance_method.py
ekopylova: Forgot to remove this line, not required.

distance-method/distance_method.py
ekopylova, line 424: I don't think so because I need this array created once for each gene (normalize_distances() is called within a loop that iterates over MSA results of each gene in the query genome).

distance-method/distance_method.py
ekopylova, line 434: I don't think so.. I need to scan the file once to determine for which species othologous genes do not exist. Then I use this information of 'missing' genes to scan through the file again and pad the distance matrix with nan's for those missing species. Basically I do this to allow indexing the full_distance_matrix by species and gene number and for genes that don't exist, the retrieved value will be nan and ignored in statistical computations (mean, std dev, Z-score).

distance-method/distance_method.py
ekopylova, line 444: True, fixed!

distance-method/distance_method.py
ekopylova, line 456: Yes!

distance-method/distance_method.py
ekopylova, line 464: Done.

distance-method/distance_method.py
ekopylova, line 489: good suggestion :) done.

distance-method/distance_method.py
ekopylova: No, I removed it.

distance-method/distance_method.py
ekopylova, line 575: Done.

distance-method/distance_method.py
antgonza, line 424: Yup, but each gene is created in preprocess_data, right? So a the end of each creation you could call normalize_distances and then this code ...

distance-method/distance_method.py
antgonza, line 434: Got it, thanks for the explanation. What about creating the full placeholder before reading and only updating the values that are in the file?

antgonza, line 434: Got it, thanks for the explanation. What about creating the full placeholder before reading and only updating the values that are in the file?

distance-method/distance_method.py
ekopylova, line 678: I need to loop over the `distance_vector` array twice: first time to fill the array and use it to compute the `mean`, `stdev`, `low_bound` and `up_bound` values, and the second time to scan the array and determine which of the cells are outside the `low_bound` and `up_bound` range. Not sure how to merge these two into one loop ..

distance-method/distance_method.py
ekopylova, line 703: Same here, the first loop is for filling the `outlier_count_matrix` matrix and the second loop is to use the completed `outlier_count_matrix` to detect outlier genes. I don't see how to do this in one step ..

distance-method/distance_method.py
ekopylova, line 434: Yup this would work and would improve the speed. I won't modify this now, or the merging of the for loops in your latter comments, since the code already runs very quickly (in comparison to other tools). I'll leave these optimizations for later (if we choose to use this method in the workflow) since there's quite a bit of other work to do.

antgonza, line 434: If you plan to leave like this, could you create an issue with all the possible improvements (mentioning the function name and the lines around it) so we do not forget? Thanks.

distance-method/distance_method.py
antgonza, line 703: It sounds like all this block could benefit from using numpy masks.

ekopylova:
@antgonza thanks for the review! I've made most of the modifications and will leave some of the suggested optimizations for later (see issue #3).

distance-method/distance_method.py
```diff
@@ -0,0 +1,957 @@
...
+    return sum(imap(operator.ne, str1, str2))
+
+
+def preprocess_data(working_dir,
+                    target_proteomes_dir,
+                    extensions,
+                    verbose=False):
+    """ Map each gene to sudo name (ex. 1_1 for species 1, gene 1).
+
+    Notes
```
antgonza: Notes should be at the end of the block ...

distance-method/distance_method.py
antgonza: K what about:
query, ref = line.split()[:2]

antgonza:
Two more minor comments and I think this is ready to go, except for https://github.com/biocore/WGS-HGT/issues/3

distance-method/distance_method.py
ekopylova: yes this works, thanks!

PR #2, opened 2015-09-10
ALF simulations, reformat input trees, launch benchmark scripts

ekopylova:
launch_software.sh : launch HGT benchmark (currently includes only phylogenetic tools)
test_1_simulate_genomes.sh : launch ALF simulations (Test 1)
reformat_input.py : reformat trees in Newick format to correspond to the input required by different phylogenetic tools
test_reformat_input.py : unit tests for reformat_input.py


benchmark/launch_software.sh
```diff
@@ -0,0 +1,199 @@
...
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+# purpose: launch HGT software on testing data sets, reformat input trees to
+#          follow input format for each tool and parse output files to report
+#          standardized statistics (number of HGTs, donors, recipients, gains
+#          and losses)
+
+# working dir
+working_dir=$1
```
antgonza, line 23: Should all these echo be within a verbose variable?

benchmark/launch_software.sh
```diff
@@ -0,0 +1,199 @@
...
+base_output_file="output_file.txt"
+hgt_summary_file=$working_dir/"hgt_summary.txt"
+input_file_nwk=$working_dir/$base_input_file_nwk
+input_file_nex=$working_dir/$base_input_file_nex
+output_file=$working_dir/$base_output_file
+input_msa_phy=$working_dir/$base_input_file_phy
+stderr=$working_dir/"stderr.txt"
+stdout=$working_dir/"stdout.txt"
+
+if [ ! -d "${working_dir}" ]; then
```
antgonza: You could use `mkdir -p` and avoid the if

benchmark/launch_software.sh
```diff
@@ -0,0 +1,199 @@
...
+                                            --gene-tree-fp $gene_tree \
+                                            --species-tree-fp $species_tree_fp \
+                                            --output-tree-fp $input_file_nwk
+    cp $input_file_nwk $trex_install_dir
+    TIME="$( time (cd $trex_install_dir; ./hgt3.4 -inputfile=$base_input_file_nwk 1>$stdout 2>$stderr) 2>&1)"
+    python ${scripts_dir}/parse_output.py --hgt-results-fp $stdout --method 'trex' >> $hgt_summary_file
+    printf "\t" >> $hgt_summary_file
+    user_time=$(echo $TIME | awk '{print $1;}')
+    wall_time=$(echo $TIME | awk '{print $2;}')
+    total_user_time_trex=$(python -c "print $total_user_time_trex + $user_time")
```
antgonza: Why use python to add to numbers vs using bash?

antgonza: This applies to all additions.

benchmark/reformat_input.py
```diff
@@ -0,0 +1,407 @@
...
+endblock;
+begin distribution;
+Range MAPPING;
+endblock;
+"""
+    # create a mapping between the species and gene tree leaves
+    mapping_dict = species_gene_mapping(gene_tree=gene_tree,
+                                        species_tree=species_tree)
+    # set branch lengths to None
+    for node in gene_tree.postorder():
```
antgonza: I guess this can be done in species_gene_mapping, right?

antgonza:
Some general comments but looks good.

benchmark/launch_software.sh
ekopylova, line 23: verbose variable created.

benchmark/launch_software.sh
ekopylova: No specific reason, changed to bash to be consistent.

benchmark/reformat_input.py
ekopylova: hmm I didn't want to because I may want to use `species_gene_mapping()` for another tool where I don't need to remove the branch lengths. However, I created a new function `remove_branch_lengths()` and moved the code there, which is also used by ranger-dtl.

PR #5, opened 2015-09-15
compute accuracy

ekopylova:
Compute TP, FP, FN, precision, recall and F-score for simulated data sets (compute_accuracy.py is called on the results of launch_sofware.sh).

### DIFF
--- a/benchmark/compute_accuracy.py
+++ b/benchmark/compute_accuracy.py
@@ -13,7 +13,7 @@
 import sys
 import click
 import re
-
+from collections import defaultdict
 
 # strings for parsing HGT information from logfile.txt of ALF simulations
 hgt_search_string = "lgt from organism "
@@ -75,16 +75,14 @@ def parse_observed_transfers(observed_hgts_f,
     1	1001	0	0	0	0	0.00 0.00
     ..
     """
-    tools = {}
     tools_id = []
+    tools = defaultdict(list)
+    # skip header line
     next(observed_hgts_f)
     for line in observed_hgts_f:
         line = line.strip().split('\t')
-        if line[0].startswith('#'):
-            for tool in line[2:]:
-                tools_id.append(tool)
-                if tool not in tools:
-                    tools[tool] = []
+        if line[0] == "#":
+            tools_id = [tool for tool in line[2:]]
             continue
         gene_id = line[1]
         for i, entry in enumerate(line[2:]):
@@ -92,7 +90,7 @@ def parse_observed_transfers(observed_hgts_f,
             # consel output (2 p-values for AU Test)
             if len(entry) == 2:
                 # species tree p-value
-                if entry[0] != "0.00":
+                if float(entry[0]) != 0.00:
                     # gene tree p-value
                     pvalue = entry[1]
                     if pvalue != 'NaN':
--- a/benchmark/compute_accuracy.py
+++ b/benchmark/compute_accuracy.py
@@ -6,15 +6,18 @@
 # The full license is in the file COPYING.txt, distributed with this software.
 # ----------------------------------------------------------------------------
 
-"""
-Given known HGTs, compute precision, recall and F-score for various tools
-=========================================================================
-"""
+#
+# Given known HGTs, compute precision, recall and F-score for various tools.
+#
 
 import sys
 import click
 import re
 
+
+# strings for parsing HGT information from logfile.txt of ALF simulations
+hgt_search_string = "lgt from organism "
+hgt_parse_string = 'lgt from organism | with gene | to organism |, now gene '
 
 def parse_expected_transfers(ground_truth_f):
     """ Parse ALF's log file.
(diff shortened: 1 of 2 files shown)


benchmark/compute_accuracy.py
```diff
@@ -0,0 +1,182 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2015--, The WGS-HGT Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+"""
```
antgonza: Comments

benchmark/compute_accuracy.py
```diff
@@ -0,0 +1,182 @@
...
+        list of transfers with each tuple representing (organism donor, gene
+        donated, organism recipient, gene received)
+
+    Notes
+    -----
+    Report horizontal gene transfers with information on organism donor,
+    gene donated, organism recipient and gene received.
+    """
+    expected_transfers = []
+    string = "lgt from organism "
```
antgonza: Should this be constants at the beginning of the file so they are easy to find?

benchmark/compute_accuracy.py
```diff
@@ -0,0 +1,182 @@
...
+    0	1000	1	1	1	1	0.96 0.04
+    1	1001	0	0	0	0	0.00 0.00
+    ..
+    """
+    tools = {}
+    tools_id = []
+    next(observed_hgts_f)
+    for line in observed_hgts_f:
+        line = line.strip().split('\t')
+        if line[0].startswith('#'):
```
antgonza: Does this if need an else or is fine to continue to the next for?

benchmark/compute_accuracy.py
```diff
@@ -0,0 +1,182 @@
...
+    -------
+    tools_accuracy: dict
+        dictionary of tuples for the number of true-positive (tp),
+        false-positive (fp), false-negative (fn) reported HGTs and precision,
+        recall and F-score
+    """
+    exp_s = set()
+    obs_s = set()
+    tools_accuracy = {}
+    for tup in expected_transfers:
```
antgonza: Normally this is nicer to have at creation: 
`exp_s = set([tup[1] for tup in expected_transfers])`

benchmark/compute_accuracy.py
ekopylova: I thought to leave the strings in separate functions since there will be more of them per tool. Currently I only have the string for parsing number of HGTs, but there will also be one for getting the gene loss count, gene duplication count and donor/recipient info. It could get messy at the top of the file if we store them all there ..

antgonza: I'm not sure about getting messy for 3 options, dozens I would agree. They could be dicts too, it it makes it clearer.

ekopylova: never mind, was looking at the wrong code. Yes we can here ..

benchmark/compute_accuracy.py
ekopylova: It's fine to continue to the next for, this `if` statement is for parsing the `#    gene ID T-REX   RANGER-DTL  RIATA-HGT   Jane 4  Consel` line to collect the names of software.

benchmark/compute_accuracy.py
ekopylova: Done.

benchmark/compute_accuracy.py
```diff
@@ -0,0 +1,182 @@
...
+
+    Notes
+    -----
+    Report horizontal gene transfers with information on organism donor,
+    gene donated, organism recipient and gene received.
+    """
+    expected_transfers = []
+    string = "lgt from organism "
+    for line in ground_truth_f:
+        if string in line:
```
wasade: could do `line.startswith('lgt from organism')`, will be more efficient

benchmark/compute_accuracy.py
```diff
@@ -0,0 +1,182 @@
...
+    """
+    tools = {}
+    tools_id = []
+    next(observed_hgts_f)
+    for line in observed_hgts_f:
+        line = line.strip().split('\t')
+        if line[0].startswith('#'):
+            for tool in line[2:]:
+                tools_id.append(tool)
+                if tool not in tools:
```
wasade: this pattern could be reduced by replacing `tools = {}` with `tools = defaultdict(list)`. You can in get `defaultdict` from `collections`

benchmark/compute_accuracy.py
```diff
@@ -0,0 +1,182 @@
...
+                    tools[tool] = []
+            continue
+        gene_id = line[1]
+        for i, entry in enumerate(line[2:]):
+            entry = entry.split()
+            # consel output (2 p-values for AU Test)
+            if len(entry) == 2:
+                # species tree p-value
+                if entry[0] != "0.00":
```
wasade: since the natural representation is a float, should this just be cast to float first?

benchmark/compute_accuracy.py
ekopylova: I'm not sure this will work because the line doesn't start with the string `lgt from organism`, the format of the log file is:
..

```
time 3.2440: lgt from organism 2 with gene 2594 to organism 6, now gene 4126
    orthologues replacement, gene 2594 in organism 6 deleted
...
where the `time` is variant (you can see an example at the bottom of `test_compute_accuracy.py`)
```

wasade: Oh, I see. Thanks!

On Wed, Sep 16, 2015 at 7:59 AM, Evguenia Kopylova <notifications@github.com

benchmark/compute_accuracy.py
ekopylova: Good suggestion, done.

benchmark/compute_accuracy.py
ekopylova: Yes, this works too, edited.

PR #6, opened 2015-09-16
parse output function

ekopylova:
Parse the output of different tools to retrieve the number of HGTs detected. The code organization allows for further development to parse for gene losses, gains and donors/recipients of HGTs.


benchmark/parse_output.py
```diff
@@ -0,0 +1,163 @@
...
+    ----------
+    input_f: string
+        file descriptor for T-REX output results
+
+    Returns
+    -------
+    number_of_hgts: string
+        number of HGTs reported by a tool, or NaN if an error occurred
+    """
+    string = 'hgt : number of HGT(s) found = '
```
antgonza: Should this be constants for easy modification?

benchmark/parse_output.py
```diff
@@ -0,0 +1,163 @@
...
+    ----------
+    input_f: string
+        file descriptor for RANGER-DTL output results
+
+    Returns
+    -------
+    number_of_hgts: string
+        number of HGTs reported by a tool, or NaN if an error occurred
+    """
+    string = "The minimum reconciliation cost is: "
```
antgonza: Same about constants.

benchmark/parse_output.py
```diff
@@ -0,0 +1,163 @@
...
+    for line in input_f:
```
antgonza: This will always only return the first line, right?

benchmark/parse_output.py
```diff
@@ -0,0 +1,163 @@
...
+
+    Returns
+    -------
+    number_of_hgts: string
+        number of HGTs reported by a tool, or NaN if an error occurred
+    """
+    string = "There are "
+    for line in input_f:
+        if string in line:
+            return line.split(string)[1].split(" component(s)")[0]
```
antgonza: This will always only return the first line, right?

benchmark/parse_output.py
```diff
@@ -0,0 +1,163 @@
...
+    ----------
+    input_f: string
+        file descriptor for RIATA-HGT output results
+
+    Returns
+    -------
+    number_of_hgts: string
+        number of HGTs reported by a tool, or NaN if an error occurred
+    """
+    string = "Host Switch: "
```
antgonza: Same comments about constants and only returning first time it founds that line.

benchmark/parse_output.py
```diff
@@ -0,0 +1,163 @@
...
+        file descriptor for Consel output results
+
+    Returns
+    -------
+    pvalues: list
+        list of P-values
+    """
+    pvalues = []
+    # skip header lines
+    input_f.next()
```
antgonza: What about put it within a for loop and add the number of lines to skip as a constant.

benchmark/parse_output.py
```diff
@@ -0,0 +1,163 @@
...
+    with open(hgt_results_fp, 'U') as input_f:
+        if method == 'ranger-dtl':
+            output = parse_rangerdtl(input_f=input_f)
+        elif method == 'trex':
+            output = parse_trex(input_f=input_f)
+        elif method == 'riata-hgt':
+            output = parse_riatahgt(input_f=input_f)
+        elif method == 'jane4':
+            output = parse_jane4(input_f=input_f)
+        elif method == 'consel':
```
antgonza, line 116: This needs and else, right? Method not implemented.

benchmark/parse_output.py
ekopylova: Currently there are only 4 parsing functions that each use a unique `string` for parsing their output files for HGTs. We will also be adding specific strings to each function for parsing gene losses, gains and donors/recipients (that's another minimum 4 strings per function). Likely, we will also be adding parsing functions for other HGT tools which will also involve unique strings for parsing HGTs. Wouldn't that be too crowded as constants at top of file?

benchmark/parse_output.py
ekopylova: If you mean that the function will return HGT info from the first occurrence of the line with `string` present, then yes. Since we're passing one gene tree for analysis, there should only be one occurrence of the line.

benchmark/parse_output.py
antgonza: OK, then what about creating a general function with this behavior and simply change the string based on tool/functions?

benchmark/parse_output.py
antgonza: Thanks for clarifying.

benchmark/parse_output.py
ekopylova: Yes, though again this line is a summary of the total HGTs found, so there should only be 1 occurrence of the line per gene (we pass only 1 gene tree at a time).

benchmark/parse_output.py
```diff
@@ -0,0 +1,163 @@
...
+        number of HGTs reported by a tool, or NaN if an error occurred
+    """
+    string = 'hgt : number of HGT(s) found = '
+    for line in input_f:
+        if string in line:
+            return line.split(string)[1].strip()
+    return "NaN"
+
+
+def parse_rangerdtl(input_f):
```
wasade: these parse methods basically look the same, would it be possible to write a single general parser?

PR #17, opened 2016-01-11
Issue 14

ekopylova:
Simulate horizontal gene transfers using 2 complete genomes (input in GenBank format).


.travis.yml
```diff
@@ -1,31 +1,50 @@
 language: python
 sudo: false
 env:
-  - PYTHON_VERSION=2.7
+  - PYTHON_VERSION="3.4"
+  - PYTHON_VERSION="3.5"
```
wasade, line 5: :)

benchmark/parse_output.py
```diff
@@ -89,6 +89,40 @@ def parse_consel(input_f):
...
+    method: string
+        tool used to detect HGTs
+
+    Returns
+    -------
+    output: string
+        number of HGTs detected
+    """
+    with open(hgt_results_fp, 'r') as input_f:
+        if (method == 'ranger-dtl' or
```
wasade, line 109: This doesn't branch much, but you might consider using a dispatch pattern?

``` python
methods = {'ranger-dtl': foo, ...}

with open(hgt_results_fp, 'r') as input_f:
    fn = methods.get(method)
    if fn is None:
        raise ValueError("blah")
```

benchmark/parse_output.py
```diff
@@ -89,6 +89,40 @@ def parse_consel(input_f):
...
+                method == 'trex' or
+                method == 'jane4' or
+                method == 'riata-hgt'):
+            output = parse_hgts(input_f=input_f,
+                                method=method)
+        elif method == 'consel':
+            output = parse_consel(input_f=input_f)
+            if output is None:
```
wasade: it seems as though the return from `parse_consel` differs from `parse_hgts` which impacts its composability as it needs a special case. Would it make sense to push this `if` statement into `parse_consel` itself?

benchmark/reformat_input.py
```diff
@@ -113,8 +113,8 @@ def species_gene_mapping(gene_tree,
         else:
             raise ValueError(
                 "Species %s does not exist in the species tree" % species)
-
-    return mapping_leaves
+    return OrderedDict(sorted(mapping_leaves.items(),
```
wasade, line 116: The docstring indicates a `list` `tuple`s, but this is an `OrderedDict`. Would it be possible to clarify the difference? Second, out of curiosity, what's the motivation for an `OrderedDict`?

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+import glob
+import random
+from operator import itemgetter
+
+from skbio import Sequence
+
+
+def extract_genbank(genbank_fp, verbose=False):
+    """Extract protein coding sequences from GenBank record.
+    Parameters
```
wasade, line 50: newline please

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+        protein IDs and the value being a 4-element list including the
+        translated sequence, the start and end positions in the genome
+    """
+    genes = {}
+    if verbose:
+        sys.stdout.write("\tParse GenBank record ...\n")
+    seq = Sequence.read(genbank_fp, format='genbank')
+    if verbose:
+        sys.stdout.write("\t\tDone.\n")
+    for feature in list(seq.interval_metadata.features.keys()):
```
wasade: Can be simplified to `for feature in seq.interval_metadata.features`

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+        if feature['type_'] == 'CDS':
+            protein_id = feature['protein_id']
+            translation = feature['translation']
+            strand = '+'
```
wasade: totally minor, but this would be a good use for a ternary:

``` python
strand = '-' if feature['rc_'] else '+'
```

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+            if feature['rc_']:
+                strand = '-'
+            loc = seq.interval_metadata.features[feature]
+            start_pos = loc[0][0]
```
wasade, line 76: could probably do the expansion inline, but i can't remember if pep8 hates this:

``` python
start_pos, end_pos = loc[0]
```

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+        # skip header lines
+        for _ in range(7):
+            next(id_pairs_f)
+        # parse orthologous family groups
+        for line in id_pairs_f:
+            line = line.strip().split()
+            # include only families with at least 2 orthologs
+            if len(line[1:-1]) > 1:
+                orthologous_groups.append(line[1:-1])
+    with open(join(results_dir, "SpeciesIDs.txt"), 'r') as species_ids_f:
```
wasade: it looks like the logic for parsing species IDs and sequence IDs is the same, would it make sense to decompose context block? eg

``` python
species_ids = _parse_orthofinder_ids(join(results_dir, 'SpeciesIDs.txt'))
sequence_ids = _parse_orthofinder_ids(join(results_dir, 'SequenceIDs.txt'))
```

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+    """Simulate orthologous replacement HGT.
+
+    Notes
+    -----
+    Algorithm: Using list of orthologous genes between donor and recipient
+    genomes
+
+    Parameters
+    ----------
+    genes_donor: dictionary
```
wasade, line 186: If possible, I think it would be useful to have 1 sentence descriptions of the parameters since it looks like this is a public method

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+    -------
+    seq_recip: skbio.sequence.Sequence
+        recipient genome sequence with HGTs
+    """
+    # number of HGTs to simulate
+    num_hgts = int(percentage_hgts*orthologous_rep_prob*len(genes_recip))
+    if num_hgts < 1:
+        num_hgts = 1
+    num_orthogroups = len(orthologous_groups)
+    idx = random.sample(list(range(0, num_orthogroups)), num_hgts)
```
wasade: should just be able to do `random.sample(range(num_orthogroups), num_hgts)`

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+    log_f.write("#type\tdonor\tstart\tend\trecipient\tnew label "
+                "recipient\tstart\tend\tstrand\n")
+    for x in range(0, num_hgts):
```
wasade: would it make sense to do `for i in idx`?

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+        orthogroup = orthologous_groups[idx[x]]
+        substitute_genes = ['*', '*']
+        # randomly select two orthologous genes from the same family
+        # representing the donor and recipient genomes
+        while '*' in substitute_genes:
+            idx2 = random.randrange(0, len(orthogroup))
```
wasade, line 243: it's possible to get repeats in this loop, will that be problematic?

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+                substitute_genes[0], substitute_genes[1]))
+        # rename recipient orthologous gene to donor's
+        hgt_gene = "%s_hgt_o" % gene_donor_label
+        genes_recip[hgt_gene] = genes_recip.pop(gene_recip_label)
+        # replace recipient gene (translated sequence) with donor's
+        genes_recip[hgt_gene][0] = genes_donor[gene_donor_label][0]
+        # update end position of HGT gene (as it can be shorter/longer than
+        # the recipient gene replaced)
+        genes_recip[hgt_gene][2] =\
+            genes_recip[hgt_gene][1] + len(genes_recip[hgt_gene][0])*3
```
wasade, line 271: why `*3`?

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+            genes_recip[hgt_gene][1:]
+        start_pos_donor, end_pos_donor, strand_donor =\
+            genes_donor[gene_donor_label][1:]
+        seq_recip = Sequence(str(seq_recip[:start_pos_recip]) +
+                             str(seq_donor[start_pos_donor:end_pos_donor]) +
+                             str(seq_recip[end_pos_recip:]))
+        if strand_recip != strand_donor:
+            genes_recip[hgt_gene][3] = genes_donor[gene_donor_label][3]
+        # write HGTs to log file
+        log_f.write("o\t%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s\n" % (
```
wasade, line 283: could condense this down with `'\t'.join(...)`

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+        start_pos_recip, end_pos_recip, strand_recip =\
+            genes_recip[hgt_gene][1:]
+        start_pos_donor, end_pos_donor, strand_donor =\
+            genes_donor[gene_donor_label][1:]
+        seq_recip = Sequence(str(seq_recip[:start_pos_recip]) +
+                             str(seq_donor[start_pos_donor:end_pos_donor]) +
+                             str(seq_recip[end_pos_recip:]))
+        if strand_recip != strand_donor:
+            genes_recip[hgt_gene][3] = genes_donor[gene_donor_label][3]
+        # write HGTs to log file
```
wasade, line 282: is this logging information that is informational for the user, or output information that is intended for a subsequent step in the workflow?

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+                             orthologous_groups,
+                             orthologous_rep_prob,
+                             percentage_hgts,
+                             log_f):
+    """Simulate orthologous replacement HGT.
+
+    Notes
+    -----
+    Algorithm: Using list of orthologous genes between donor and recipient
+    genomes
```
wasade: Possible to expand this slightly?

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+                       genes_recip,
+                       seq_recip,
+                       orthologous_rep_prob,
+                       percentage_hgts,
+                       log_f):
+    """Simulate novel gene acquisition HGT.
+
+    Parameters
+    ----------
+    genes_donor: dictionary
```
wasade, line 306: same comment here

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+    Returns
+    -------
+    seq_recip: skbio.sequence.Sequence
+        recipient genome sequence with HGTs
+
+    Notes
+    -----
+    Algorithm:
+        1. choose random location in recipient genome where to insert a gene
+           (randomly chosen from list of donor genes)
```
wasade: ~~randomly~~

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+    log_f: file descriptor
+
+    Returns
+    -------
+    seq_recip: skbio.sequence.Sequence
+        recipient genome sequence with HGTs
+
+    Notes
+    -----
+    Algorithm:
```
wasade, line 328: This is very nice, thank you

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+           (randomly chosen from list of donor genes)
+        2. use gene (recipient) positioning array to locate an open region
+           (that doesn't include an existing gene) near the random location
+           to insert the new gene (we want to avoid gene overlap so that
+           compositional methods can clearly pick out individual coding
+           genes)
+        3. insert new gene, record existance in gene positioning array
+    """
+    num_hgts = int(percentage_hgts*(1-orthologous_rep_prob)*len(genes_recip))
+    if num_hgts < 1:
```
wasade: `num_hgts = max(1, num_hgts)` ?

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+        num_hgts = 1
+    # create recipient genome gene positioning array
```
wasade: would it be possible to expand this logic out? something like:

``` python
gene_positions = [(0, 0), (len(seq_recip), len(seq_recip))]
for _, start, stop in gene_recip.values():
    gene_positions.append((start, stop))
```

...that assumes I'm interpreting this accurately though...

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+    gene_positions =\
+        [(genes_recip[gene][1], genes_recip[gene][2]) for gene in genes_recip]
+    # add start and end positions of recipient genome to allow for HGTs
+    # simulated before the first and after the last existing gene
+    gene_positions.append((0, 0))
+    gene_positions.append((len(seq_recip), len(seq_recip)))
+    # sort array for gene positions in ascending order
+    gene_positions_s = sorted(gene_positions, key=itemgetter(0))
+    # select a random list of positions where to insert the new gene
+    idx = random.sample(list(range(0, len(gene_positions_s)-1)), num_hgts)
```
wasade: no need for `list`

wasade: also no need for `0`

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+    gene_donor_labels = random.sample(list(genes_donor.keys()), num_hgts)
+    log_f.write("#type\tdonor\tstart\tend\trecipient\tstart\t"
```
wasade, line 352: same comment here about whether this is informational or part of the pipeline

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+        [(genes_recip[gene][1], genes_recip[gene][2]) for gene in genes_recip]
+    # add start and end positions of recipient genome to allow for HGTs
+    # simulated before the first and after the last existing gene
+    gene_positions.append((0, 0))
+    gene_positions.append((len(seq_recip), len(seq_recip)))
+    # sort array for gene positions in ascending order
+    gene_positions_s = sorted(gene_positions, key=itemgetter(0))
+    # select a random list of positions where to insert the new gene
+    idx = random.sample(list(range(0, len(gene_positions_s)-1)), num_hgts)
+    gene_donor_labels = random.sample(list(genes_donor.keys()), num_hgts)
```
wasade: no need for `keys()`

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+    log_f.write("#type\tdonor\tstart\tend\trecipient\tstart\t"
+                "end\tstrand\n")
+    # begin simulation
+    for x in range(0, num_hgts):
```
wasade: no need for `0`

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+        num_hgts = 1
+    # create recipient genome gene positioning array
+    gene_positions =\
+        [(genes_recip[gene][1], genes_recip[gene][2]) for gene in genes_recip]
+    # add start and end positions of recipient genome to allow for HGTs
+    # simulated before the first and after the last existing gene
+    gene_positions.append((0, 0))
+    gene_positions.append((len(seq_recip), len(seq_recip)))
+    # sort array for gene positions in ascending order
+    gene_positions_s = sorted(gene_positions, key=itemgetter(0))
```
wasade: I don't think `itemgetter` is needed here since the sort should default to the first index

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+    # begin simulation
+    for x in range(0, num_hgts):
+        # select random donor gene (for HGT)
+        gene_donor_label = gene_donor_labels[x]
+        idx_recip = gene_positions_s[idx[x]][1] + 1
+        # beginning from valid position for inserting new gene, check whether
+        # the length can fit without overlapping with existing gene, otherwise
+        # search for next valid position
+        for y in range(idx[x], len(gene_positions_s)-1):
+            if idx_recip + len(genes_donor[gene_donor_label][0])*3 <\
```
wasade, line 363: why `*3`? 

`idx_recip + len(genes_donor[gene_donor_label[0]) * 3` can be computed outside of the nested loop, and would reduce the semantic complexity of the conditional

wasade, line 363: oh... sorry, can't include `idx_recip`

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+    # select a random list of positions where to insert the new gene
+    idx = random.sample(list(range(0, len(gene_positions_s)-1)), num_hgts)
+    gene_donor_labels = random.sample(list(genes_donor.keys()), num_hgts)
+    log_f.write("#type\tdonor\tstart\tend\trecipient\tstart\t"
+                "end\tstrand\n")
+    # begin simulation
+    for x in range(0, num_hgts):
+        # select random donor gene (for HGT)
+        gene_donor_label = gene_donor_labels[x]
+        idx_recip = gene_positions_s[idx[x]][1] + 1
```
wasade, line 358: should this be `x + 1` given the increment on line 348?

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+    # output dir for simulated results
+    simulated_dir = join(output_dir, "simulated")
+    if not exists(simulated_dir):
+        makedirs(simulated_dir)
+    donor_genome_aa_fp = join(
+        simulated_dir, "%s.faa" % basename(splitext(donor_genome_fp)[0]))
+    recip_genome_aa_fp = join(
+        simulated_dir, "%s.faa" % basename(splitext(recip_genome_fp)[0]))
+    with open(donor_genome_aa_fp, 'w') as donor_genome_aa_f:
+        for gene in genes_donor:
```
wasade, line 435: why are these manually formatted whereas the nucleotide data is using the skbio FASTA formatter?

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+    recip_genome_nucl_fp = join(
+        simulated_dir, "%s.fna" % basename(splitext(recip_genome_fp)[0]))
+    seq_donor.write(donor_genome_nucl_fp, format='fasta')
+    seq_recip.write(recip_genome_nucl_fp, format='fasta')
+
+    return (donor_genome_nucl_fp, donor_genome_aa_fp, recip_genome_nucl_fp,
+            recip_genome_aa_fp)
+
+
+def simulate_hgts(seq_donor,
```
wasade, line 453: i think most of these parameters are not used by the method directly, but are instead passed to methods. Would it make sense to collect in `**kwargs` and pass that on?

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+        if int(day) < 10:
+            day = "0%s" % day
+        results_dir = join(
+            proteomes_dir, "Results_%s%s" % (date[1], day),
+            "WorkingDirectory")
+        species_ids, sequence_ids, orthologous_groups =\
+            parse_orthofinder(results_dir)
+        # no orthologs found, exit orthologous replacement simulation
+        if orthologous_groups:
+            seq_recip = simulate_orthologous_rep(genes_donor,
```
wasade, line 511: `seq_recip` is reused here from the input parameters, is that expected? It also seems to be reused on line 503

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,627 @@
...
+        recip_genome_fp=recipient_genbank_fp,
+        output_dir=output_dir,
+        percentage_hgts=percentage_hgts,
+        orthologous_rep_prob=orthologous_rep_prob,
+        log_f=log_f,
+        threads=threads,
+        verbose=verbose)
+
+
+@click.command()
```
wasade, line 603: There's a logical separation here as this is a user interface method, would it make sense to partition this file into "business logic" and "UI logic"?

wasade, line 603: ...well, the UI method is rather small and only a single one, so perhaps not..

benchmark/tests/test_simulate_hgts.py
```diff
@@ -0,0 +1,549 @@
...
+            "TTGATTTTAGACTGGTATTTCAAGAACGATTACTTTTAAACTGGCGTTTAAAT"
+            "ATCAACATCTCCCAGCTATCCTACACAAAAA")
+
+    def tearDown(self):
+        rmtree(self.working_dir)
+
+    def test_extract_genbank(self):
+        """Test parsing sequence and gene information from a GenBank record.
+        """
+        print("test_extract_genbank")
```
wasade: possible to remove prints from test code?

wasade: ...using `-v` when executing will show what method is being run

benchmark/reformat_input.py
ekopylova, line 116: the doc string should be OrderedDict, will update. The reason for OrderedDict is so that when I output the dictionary to a file the entries are always in the same order, easier for debugging and testing ..

benchmark/simulate_hgts.py
ekopylova: yes, good suggestion!

benchmark/simulate_hgts.py
ekopylova, line 243: I added a inline note on this, however:

Each orthogroup is guranteed to have at least two genes, one from donor (prefixed with '0') and second from recipient (prefixed with '1'). The following while loop will continue until an index for two genes prefixed with '0' and '1' is selected. At each iteration the chances the while loop must continue reduce exponentially since the index is chosen randomly form the same set of options.

So, there can be repeats but it shouldn't loop for too long given that we're looking at orthologous families between two genomes and often the size of the orthogroup will be 2.

benchmark/simulate_hgts.py
ekopylova, line 271: codon.

We need to translate codons into nucleotides in order to insert the gene into the respective genome.

Added a note in the code, thanks!

benchmark/simulate_hgts.py
ekopylova, line 282: Information for the user. It lists info regarding simulated HGTs.

benchmark/simulate_hgts.py
ekopylova: `gene_positions = [(0, 0), (len(seq_recip), len(seq_recip))]` is a good suggestion. However, the values of `gene_recip` are actually lists containing 4 variables - the gene sequence, start position, end position and the strand (+ or -).

ekopylova: `for seq, start, stop, strand in genes_recip.values():`` worked

benchmark/simulate_hgts.py
ekopylova, line 435: `genes_donor` and `genes_recip` are dictionaries whereas `seq_donor` and `seq_recip` are skbio Sequences ..

benchmark/simulate_hgts.py
ekopylova, line 453: when I pass `**kwargs` should I keep the Parameters section the same?

ekopylova, line 453: and to subsequent functions pass `kwargs[seq_donor]` and so on ..

benchmark/simulate_hgts.py
ekopylova, line 511: Yes, that's correct. I'm editing `seq_recip` from it's original form to one with simulated HGTs (first orthologous replacement and then novel acquisition), then I write the final version to file.

benchmark/simulate_hgts.py
ekopylova, line 363: `*3` to expanded nucleotide length from codons

benchmark/simulate_hgts.py
wasade, line 453: One strategy is to allow the called functions to accept `**kwargs`, so then you just pass it on. for instance:

``` python
def foo(some, parameter, **voodoo):
    pass

def bar(stuff, **voodoo):
    pass

def controller(important, controller, stuff, **voodoo):
    # ...
    foo(x, y, **voodoo)
    # ...
    bar(stuff, **voodoo)
```

ekopylova, line 453: Hmm, I'm using 6/12 parameters (output_dir, donor_genome_fp, recip_genome_fp, genes_donor, genes_recip, orthologous_rep_prob) passed to `foo` in `foo`. I'll need to unpack them with **kwargs, does it still make sense to use it?

wasade, line 453: Expansion can be done implicitly:

``` python
12:15:04 (daniel@sandbar):~> def foo(a, b, c):
   ...:     return (a+b+c)
   ...:

12:15:17 (daniel@sandbar):~> foo(**{'a':1, 'b': 2, 'c': 3})
Out[2]: 6

12:15:52 (daniel@sandbar):~> foo(10, **{'c': 2, 'b': 3})
Out[4]: 15
```

wasade:
:+1:

.travis.yml
```diff
@@ -1,33 +1,52 @@
...
 before_install:
   - wget http://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
   - chmod +x miniconda.sh
   - bash miniconda.sh -b -p $HOME/miniconda
   - export PATH="$HOME/miniconda/bin:$PATH"
   - conda config --set always_yes yes --set changeps1 no
   - conda config --add channels https://conda.anaconda.org/biocore
+  - conda config --add channels https://conda.anaconda.org/mortonjt
```
antgonza, line 13: Just out of curiosity, what's the plan for this? Is this ever gonna move to biocore?

benchmark/simulate_hgts.py
```diff
@@ -0,0 +1,669 @@
...
+
+
+def _parse_orthofinder_ids(ids_fp):
+    """
+    """
+    ids = {}
+    with open(ids_fp, 'r') as ids_f:
+        for line in ids_f:
+            line = line.strip().split()
+            ids[line[0].split(':')[0]] = line[1]
```
antgonza: Just want to make sure we are in the same page. This means that if we have 2 ids like: `bla: foo1` and `bla: foo2` you will discard/overwrite one of them, right?

antgonza:
A couple of questions.

benchmark/simulate_hgts.py
ekopylova: that would be the case, however ids in the first column are unique (and if they weren't, the ids in the second column would have to be identical for the two non-unique ids in column 1). I added though a check for unique IDs.

.travis.yml
ekopylova, line 13: Yes, eventually. First the [RFC](https://github.com/biocore/scikit-bio-rfcs/pull/9) needs to go through ..

antgonza, line 13: Got it, could you add an issue in this repo so we don't forget? Thanks.

benchmark/tests/test_simulate_hgts.py
```diff
@@ -132,8 +132,12 @@ def test_launch_orthofinder(self):
         """
         launch_orthofinder(self.proteomes_dir, 1, verbose=True)
         date = time.strftime("%c").split()
+        day = date[2]
+        if int(day) < 10:
+            day = "0%s" % day
         results_dir = join(
-            self.proteomes_dir, "Results_%s%s" % (date[1], date[2]))
+            self.proteomes_dir, "Results_%s%s" % (date[1], day))
+        print(results_dir)
```
antgonza: I guess this print can go ...

.travis.yml
ekopylova, line 13: done!

PR #18, opened 2016-01-14
Update benchmark layout, add main launch script

ekopylova:
1. Update benchmark to call individual scripts for launching software (rather than all in one call)
2. Use scikit-bio development version that supports TabularMSA and soon GenBank parsing


benchmark/launch_benchmark.sh
```diff
@@ -0,0 +1,116 @@
...
+    echo "blast nr: ${blast_nr}"
+    echo "PhyloNet install dir: $phylonet_install_dir"
+    echo "Jane 4 install dir: $jane_install_dir"
+    echo "T-REX install dir: $trex_install_dir"
+fi
+
+mkdir -p "${working_dir}/diamond"
+filename=$(basename "${query_species_coding_seqs_fp}")
+diamond_output=${working_dir}/diamond/$filename
+## launch DIAMOND if alignments don't exist for genome
```
antgonza: Remove if not needed.

benchmark/launch_benchmark.sh
```diff
@@ -0,0 +1,116 @@
...
+#                   --sensitive
+#    # convert output to tab delimited format
+#    diamond view --daa ${diamond_output}.daa -f tab -o ${diamond_output}.m8
+#fi
+
+## run GeneMarkS training (generate typical and atypical gene models)
+mkdir -p "${working_dir}/hmm-models"
+filename=$(basename "${species_genome_fp}")
+hmm_output=${working_dir}/hmm-models/$filename
+#if [ "${species_model_fp}" == "None" ]
```
antgonza: Remove?

benchmark/launch_software.sh
```diff
@@ -29,196 +29,123 @@ species_tree_fp=$3
...
-    i=$((i+1))
-done
-
-echo "Total time T-REX: $total_user_time_trex"
-echo "Total time RANGER-DTL: $total_user_time_rangerdtl"
-echo "Total time RIATA-HGT: $total_user_time_riatahgt"
-echo "Total time Jane 4: $total_user_time_jane"
-echo "Total time AU-Test: $total_user_time_consel"
+## run T-REX
+echo "source ~/.bash_profile; \
```
antgonza: This is only required for barnacle, right? Should it be something defined at the top of the file and appended to all commands?

benchmark/launch_software.sh
```diff
@@ -29,196 +29,123 @@ species_tree_fp=$3
...
+                                      ${hgt_summary}.trex.txt \
+                                      ${verbose}\
+                                      ${stdout}.trex.txt \
+                                      ${stderr}.trex.txt \
+                                      ${scripts_dir} \
+                                      ${species_tree_fp} \
+                                      ${input_file_nwk}.trex.txt \
+                                      ${trex_install_dir} \
+                                      ${base_input_file_nwk}.trex.txt" | qsub $qsub -N run_trex; sleep 2
+
```
antgonza, line 131: If the previous comment is follow, this should be in a prepend.

benchmark/run_jane4.sh
```diff
@@ -0,0 +1,53 @@
...
+scripts_dir=$6
+species_tree_fp=$7
+input_file_nex=$8
+output_file=$9
+jane_install_dir=${10}
+
+total_user_time_jane="0.0"
+total_wall_time_jane="0.0"
+i=0
+printf "#JANE4\n" >> $output_fp
```
antgonza, line 27: This file gets removed at the end, so it can be temporal, right? So no need to pass it. Also it could be a text variable, right?

benchmark/run_ranger.sh
```diff
@@ -0,0 +1,51 @@
...
+                                            --species-tree-fp $species_tree_fp \
+                                            --output-tree-fp $input_file_nwk
+    TIME="$( time (ranger-dtl-U.linux -i $input_file_nwk -o ${output_file} 1>>$stdout 2>>$stderr) 2>&1)"
+    python ${scripts_dir}/parse_output.py --hgt-results-fp ${output_file} --method 'ranger-dtl' >> $output_fp
+    printf "\n" >> $output_fp
+    user_time=$(echo $TIME | awk '{print $1;}')
+    wall_time=$(echo $TIME | awk '{print $2;}')
+    total_user_time_rangerdtl=$(echo $total_user_time_rangerdtl + $user_time | bc)
+    total_wall_time_rangerdtl=$(echo $total_wall_time_rangerdtl + $wall_time | bc)
+    rm ${output_file}
```
antgonza, line 49: Same comment about output_file ...

benchmark/run_riatahgt.sh
```diff
@@ -0,0 +1,50 @@
...
+printf "#RIATAHGT\n" >> $output_fp
+
+# search for HGTs in each gene tree
+for gene_tree in $gene_tree_dir/*.nwk
+do
+    gene_tree_file=$(basename $gene_tree)
+    gene_number=$(echo $gene_tree_file | sed 's/[^0-9]*//g')
+    printf "$i\t$gene_number\t" >> $output_fp
+
+    python ${scripts_dir}/reformat_input.py --method 'riata-hgt' \
```
antgonza, line 37: The reason that you have to run them like this (prepending python) is because they are not executable in the repo. Perhaps it's worth changing.

benchmark/run_trex.sh
```diff
@@ -0,0 +1,53 @@
...
+                                            --output-tree-fp $input_file_nwk
+    cp $input_file_nwk $trex_install_dir
+    TIME="$( time (cd $trex_install_dir; ./hgt3.4 -inputfile=${base_input_file_nwk} 1>$stdout 2>>$stderr) 2>&1)"
+    python ${scripts_dir}/parse_output.py --hgt-results-fp $stdout --method 'trex' >> $output_fp
+    printf "\n" >> $output_fp
+    user_time=$(echo $TIME | awk '{print $1;}')
+    wall_time=$(echo $TIME | awk '{print $2;}')
+    total_user_time_trex=$(echo $total_user_time_trex + $user_time | bc)
+    total_wall_time_trex=$(echo $total_wall_time_trex + $wall_time | bc)
+    rm $stdout
```
antgonza, line 51: Same about stdout

antgonza:
A few comments.

benchmark/launch_benchmark.sh
ekopylova: No, it's needed, forgot to uncomment ..

benchmark/launch_benchmark.sh
ekopylova: Same here .. this preprocessing is required if an HMM model doesn't yet exist.

benchmark/launch_software.sh
ekopylova: You're right, fixed.

benchmark/run_jane4.sh
ekopylova, line 27: I've updated the code to concatenate the output of that file after each gene to a global output file, whose prefix name is the same. With this change, it makes sense to pass the file ..

benchmark/run_riatahgt.sh
ekopylova, line 37: Decided to keep this for the time being in order to avoid duplicate script names already existing in a user's $PATH.

PR #30, opened 2016-06-17
run HGTector and parse output

qiyunzhu:
Edited parse_output.py and run_hgtector.sh to call HGTector and return a table of HGTs


ekopylova:
Can you remove lines 17-23 in .travis.yml file to resolve the KALIGN download issue? We will create a conda recipe for it in the Phylomizer branch.

.travis.yml
```diff
@@ -14,13 +14,6 @@ before_install:
   - conda update -q conda
   - conda info -a
   # Install KALIGN
-  - mkdir kalign; cd kalign
-  - wget http://msa.sbc.su.se/downloads/kalign/current.tar.gz
-  - tar zxvf current.tar.gz
-  - ./configure
-  - make
-  - export PATH=$PWD/kalign:$PATH
-  - cd ..
```
ekopylova, line 23: will add this as a conda recipe later (needed for phylomizer in my PR)

benchmark/parse_output.py
```diff
@@ -117,6 +117,32 @@ def parse_darkhorse(input_f, low_lpi=0.0, high_lpi=0.6, return_genomes=False):
...
+    reading = 0
+    output = []
+    for line in input_f:
+        line = line.rstrip('\r\n')
+        if not line:
+            continue
+        elif line.startswith('Putatively HGT-derived genes:'):
+            reading = 1
+        elif reading:
+            output.append(line.split('\t')[0])
```
ekopylova: are you returning here only the gene ID or gene ID + tax_id + taxonomy string?

benchmark/run_hgtector.sh
```diff
@@ -68,21 +71,33 @@ then
...
+    cp -f $hgtector_config_file ${working_dir}/hgtector/config.txt
 fi
 
 perl ${hgtector_install_dir}/HGTector.pl ${working_dir}/hgtector
-## print a list of predicted HGT-derived genes
-echo ${working_dir}/hgtector/result/HGT/id.txt
+
+## print predicted HGTs
+echo ""
+echo Putatively HGT-derived genes:
```
ekopylova: At this point we want to call `parse_output.py` (e.g., passing `${working_dir}/hgtector/result/detail/id.txt` and running the function `parse_hgtector()`) rather than parsing directly in the bash script.

ekopylova:
@qiyunzhu a few comments. Is it possible to also update the call to `run_hgtector.sh` from `launch_software.sh` (see [here](https://github.com/biocore/WGS-HGT/blob/master/benchmark/launch_software.sh#L236))? You only need to update all the parameters that you expect for `run_hgtector.sh`

benchmark/parse_output.py
qiyunzhu: Just Gene IDs. Previously I had the code return all fields. In order to simplify the unit test I trimmed that down to gene IDs only (so that I didn't have to append a long output_exp value in the code).

PR #31, opened 2016-07-25
update egid

qiyunzhu:


.travis.yml
```diff
@@ -54,7 +54,7 @@ before_install:
   # Install Java goodies
   - mkdir java-files
   # Install PhyloNet
-  - wget http://bioinfo.cs.rice.edu/sites/g/files/bxs266/f/kcfinder/files/PhyloNet_3.5.7.jar -O java-files/PhyloNet_3.5.7.jar
+  # wget http://bioinfo.cs.rice.edu/sites/g/files/bxs266/f/kcfinder/files/PhyloNet_3.5.7.jar -O java-files/PhyloNet_3.5.7.jar
```
antgonza: Why do you need this change? And if it's not needed, please remove.

benchmark/parse_output_gi.py
```diff
@@ -0,0 +1,79 @@
...
+    genes = {}
+    gb = Sequence.read(genbank_fp, format='genbank')
+    for feature in gb.interval_metadata.features:
+        if feature['type_'] == 'CDS':
+            if 'protein_id' in feature:
+                protein_id = feature['protein_id'].replace('\"', '')
+                loc = gb.interval_metadata.features[feature]
+                start = loc[0][0] + 1
+                end = loc[0][1]
+                if protein_id not in genes:
```
antgonza, line 40: Is it OK to discard duplicates?

benchmark/parse_output_gi.py
```diff
@@ -0,0 +1,79 @@
...
+                    genes[protein_id] = [start, end]
```
antgonza: Could this be a tuple?

benchmark/parse_output_gi.py
antgonza, line 40: Also, it will be good to have this test 3 lines above: no need to do the loc/start/end if we are not going to use it.

benchmark/parse_output_gi.py
```diff
@@ -0,0 +1,79 @@
...
+    output: string
+        gene names (protein_ids) separated by newline
+    """
+    genes = {}
+    gb = Sequence.read(genbank_fp, format='genbank')
+    for feature in gb.interval_metadata.features:
+        if feature['type_'] == 'CDS':
+            if 'protein_id' in feature:
+                protein_id = feature['protein_id'].replace('\"', '')
+                loc = gb.interval_metadata.features[feature]
```
antgonza: Just out of curiosity, will loc == feature?

benchmark/parse_output_gi.py
```diff
@@ -0,0 +1,79 @@
...
+                start = loc[0][0] + 1
+                end = loc[0][1]
+                if protein_id not in genes:
+                    genes[protein_id] = [start, end]
+    genes_in_gi = {}
+    with open(gi_fp, 'r') as input_f:
+        for line in input_f:
+            l = line.strip().split()
+            if len(l) < 2:
```
antgonza, line 49: Can you add a comment of why this might happen?

benchmark/parse_output_gi.py
```diff
@@ -0,0 +1,79 @@
...
+                continue
+            (start, end) = (int(l[0]), int(l[1]))
+            for (gene, l) in genes.items():
+                if (l[0] >= start and l[1] <= end):
+                    if gene not in genes_in_gi:
+                        genes_in_gi[gene] = 1
+    output = []
+    for gene in sorted(genes_in_gi):
```
antgonza: The next 3 lines can be replaced by:
`return '\n'.join(sorted(genes_in_gi))`
right?

benchmark/reformat_input_gi.py
```diff
@@ -0,0 +1,143 @@
...
+    nucl_seq = ''
+    genes = {}
+    nseq = 0  # number of nucleotide sequences
+    with open(genbank_fp, 'r') as input_f:
+        for line in input_f:
+            if line.startswith('//'):
+                nseq += 1
+    abs_pos = 0  # absolute position in concantenated nucleotide sequence
+    for i in range(nseq):
+        gb = Sequence.read(genbank_fp, seq_num=i+1, format='genbank')
```
antgonza, line 48: Perhaps I'm missing something but will it be easier to remove lines 38->48 and simply do:

```
for gb in Sequence.read(genbank_fp, format='genbank'):
```

? I think it will be much faster too.

benchmark/reformat_input_gi.py
```diff
@@ -0,0 +1,143 @@
...
+        locus_name = gb.metadata['LOCUS']['locus_name']
+        size = gb.metadata['LOCUS']['size']
+        loci.append([locus_name, size])
+        nucl_seq += str(gb)
+        for feature in gb.interval_metadata.features:
```
antgonza, line 53: Perhaps this can become a function that you pass the 'type_' and feature: 'translation'/etc ...

benchmark/reformat_input_gi.py
```diff
@@ -0,0 +1,143 @@
...
+"""
+
+import click
+from os import remove
+from os.path import join
+
+from skbio import Sequence, DNA
+
+
+def reformat_egid(genbank_fp,
```
antgonza, line 22: A general question: Can you use the skbio parsers and writers?

antgonza:
A few questions/comments.

.travis.yml
ekopylova: We removed it because of new issues regarding certificates and will install with conda later ([new issue](https://github.com/biocore/WGS-HGT/issues/33)). However, adding the flag `--no-check-certificate` temporarily resolves the certificate issue, @qiyunzhu could you uncomment the install command and add that flag?

qiyunzhu: Edited.

benchmark/parse_output_gi.py
qiyunzhu: Good point! I moved this test 3 lines upper. I think CDS.protein_id should be unique in a valid GenBank file. Plus, it has to be unique in our analysis. This test is to avoid some unexpected cases.

benchmark/parse_output_gi.py
qiyunzhu: I think it could. I changed it.

benchmark/parse_output_gi.py
qiyunzhu: I inherited this code from previous codes and honestly don't know why it is phrased like this. Besides, I noticed that these expressions are different in scikit-bio 0.4.2 versus 0.5.0. The latter looks more reasonable to me.

benchmark/parse_output_gi.py
qiyunzhu: In my tests, the standard EGID output files always read `<start><\t><end><\n>`. Therefore `len(l)` always `== 2`. I am thinking that, if the user wants to add a comment in this file (e.g., `# GI no. 1<\n>`), or mistakenly added some new lines or whatever. This code will make the process safer.

benchmark/parse_output_gi.py
qiyunzhu: Makes sense! Edited.

benchmark/reformat_input_gi.py
qiyunzhu: I think `Sequence.read()` without `seq_num` only reads the first record. Refer to line 314-317 of [genbank.py](https://github.com/biocore/scikit-bio/blob/master/skbio/io/format/genbank.py).

benchmark/reformat_input_gi.py
qiyunzhu: I tried to use skbio to write a GenBank format file. However, EGID had problem parsing this file. I thought part of the reason is that skbio's GenBank file has different leading space pattern comparing to native NCBI GenBank format. For example: 5 spaces + "gene" + 12 spaces is the native pattern, but skbio does not precisely replicate that (I cannot recall how it looks like).

benchmark/reformat_input_gi.py
qiyunzhu: When I wrote these codes I was inspired by some previous [codes](https://github.com/biocore/WGS-HGT/blob/master/benchmark/simulate_hgts.py#L47). Therefore you may find that the style is similar. Here is s question if we make this a function: a global variable `genes` is outside the loop. If it is an independent function, it will have to access `genes` every iteration. I guess that it may be complicated. I am not very familiar with Python's namespace and memory allocation policies. So please correct me if I am not getting this right.

qiyunzhu:
@antgonza Thank you for reviewing the code I wrote! It was very careful and inspiring. Sorry for the late response, as I wasn't keeping an eye on the repo in the past few days. @ekopylova Thank you for reminding me! I am but an amateur Python coder. So I am doing my best to mimic how you guys are writing. Previously I just looked for a random way to get things done, but your comments are gradually making me aware of how to do things correctly and beautifully. Thank you!

ekopylova:
@qiyunzhu thanks for making the changes! Could you resolve the branch conflict and push your code again? Thanks!

qiyunzhu:
Hello Jenya, no problem! What do people usually do to resolve branch
conflicts? Should I pull the latest biocide/wgs-hgt master to my local git
repository, then push to my qiyunzhu/wgs-hgt egid? Will that overwrite my
local edition? Thanks!

ekopylova:
@qiyunzhu you should see "Use the command line to resolve conflicts before continuing" at the place on the page where it says "This branch has conflicts that must be resolved". The command line instructions give you details on how to make the changes. Could you follow those? Thanks!

qiyunzhu:
@ekopylova On my side there is no "Use the command line to resolve conflicts before continuing". However I have found some tutorials to do this. I am working to meet a deadline at this moment. But I will tackle this tomorrow.

qiyunzhu:
@ekopylova Conflict solved! Thanks for reminding me!

PR #32, opened 2016-08-31
Prototype selection exhaustive

sjanssen2:
First part of a series of pull requests. I have a bunch of heuristic implementations for the "prototype selection problem", starting with the exhaustive algorithm. Later PRs will contain the mentioned heuristics.
Our use case is the selection of ~50 genomes for those species where we have >> 50 genomes as a pre-filtering step before building a large phylogenetic tree.


josenavas:
@sjanssen2 is this ready for review?

sjanssen2:
Hi Jose,

I think so, but since this is my first real pull request for sources I 
wrote on my own, I am not completely sure if I am following our normal 
workflow. So please be patient with me :-)

I appreciate every comment on my code!

Since I heart today how successful small PRs are, I am going the same 
route ... which means that there is no fancy algorithm yet. Just the 
very basics. I am going to add more advanced things with the next PRs.

Thanks,
Stefan

.travis.yml
```diff
@@ -54,7 +54,7 @@ before_install:
   # Install Java goodies
   - mkdir java-files
   # Install PhyloNet
-  - wget http://bioinfo.cs.rice.edu/sites/g/files/bxs266/f/kcfinder/files/PhyloNet_3.5.7.jar -O java-files/PhyloNet_3.5.7.jar
+  - wget http://bioinfo.cs.rice.edu/sites/g/files/bxs266/f/kcfinder/files/PhyloNet_3.5.7.jar -O java-files/PhyloNet_3.5.7.jar --no-check-certificate
```
mortonjt, line 57: Note that this was required, since the travis build was failing.  @ekopylova have you noticed this before?

phylogeny/prototypeSelection.py
```diff
@@ -0,0 +1,106 @@
...
+    Fuzzy Systems, 2008. FUZZ-IEEE 2008.
+"""
+
+from skbio.stats.distance import DistanceMatrix
+import itertools
+import numpy as np
+import scipy as sp
+
+
+def distanceSum(elements, dm: DistanceMatrix):
```
mortonjt: PY3 types ftw!

phylogeny/prototypeSelection.py
```diff
@@ -0,0 +1,106 @@
...
+    -------
+    float:
+        the sum of all pairwise distances of dm for IDs in elements
+    '''
+
+    # some assertions
+    if type(dm) != DistanceMatrix:
+        raise TypeError('dm is not of type "DistanceMatrix"')
+    assert(len(elements) > 1)
+    assert(len(elements) <= len(dm.ids))
```
mortonjt: Just out of curiosity, why is this required?

phylogeny/prototypeSelection.py
```diff
@@ -0,0 +1,106 @@
...
+def distanceSum(elements, dm: DistanceMatrix):
+    '''Compute the sum of pairwise distances for the given elements according to
+    the given dissimilarity matrix.
+
+    Parameters
+    ----------
+    elements: list or set
+        list or set of element IDs for which the sum of distances is computed
+    dm: skbio.stats.distance.DistanceMatrix
+        pairwise dissimilarity matrix.
```
mortonjt: Distance matrix or a dissimilarity matrix?

If it is a dissimilarity matrix, that implies that the symmetric property is not required.  You can actually use the [`skbio.DissimiliarityMatrix`](https://github.com/biocore/scikit-bio/blob/master/skbio/stats/distance/_base.py#L44) if this is really the case.

phylogeny/prototypeSelection.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def distanceSum(elements, dm: DistanceMatrix):
+    '''Compute the sum of pairwise distances for the given elements according to
+    the given dissimilarity matrix.
+
+    Parameters
+    ----------
+    elements: list or set
+        list or set of element IDs for which the sum of distances is computed
```
mortonjt: What exactly are element IDs?  Are they integer indices? Or string IDs?

phylogeny/prototypeSelection.py
```diff
@@ -0,0 +1,106 @@
...
+    # some assertions
+    if type(dm) != DistanceMatrix:
+        raise TypeError('dm is not of type "DistanceMatrix"')
+    assert(len(elements) > 1)
+    assert(len(elements) <= len(dm.ids))
+    if len(set(elements)) != len(elements):
+        raise Exception('List of elements contain duplicates!')
+
+    # actual computation
+    return sum([dm[pair] for pair in set(itertools.combinations(elements, 2))])
```
mortonjt: If it is really a distance matrix, this can significantly sped up by summing the upper or lower triangular matrix via [`numpy.tril`](http://docs.scipy.org/doc/numpy/reference/generated/numpy.tril.html)

``` python
In [1]: from skbio.stats.distance import randdm
In [2]: import numpy as np
In [3]: dm = randdm(100)
In [4]: elements = range(100)
In [5]: %timeit sum([dm[pair] for pair in set(itertools.combinations(elements, 2))])
100 loops, best of 3: 8.89 ms per loop

In [6]: %timeit np.tril(dm.data).sum()
The slowest run took 6.49 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 36.3 µs per loop

```

phylogeny/prototypeSelection.py
```diff
@@ -0,0 +1,106 @@
...
+
+    Parameters
+    ----------
+    distanceMatrix: skbio.stats.distance.DistanceMatrix
+        pairwise distances for all elements in the full set S.
+        Must be symmetric and non-hollow.
+    numPrototypes: int
+        Number of prototypes to select for dissimilarity matrix.
+        Must be >= 2, since a single prototype is useless.
+        Must be <= |distanceMatrix|, otherwise no reduction is neccessary.
```
mortonjt: neccessary -> necessary

mortonjt: what is `|distanceMatrix|`?  Is that the total number of elements in `distanceMatrix`?

phylogeny/prototypeSelection.py
```diff
@@ -0,0 +1,106 @@
...
+    maxCombinationsToTest: int
+        The maximal number of combinations to test. If exceeding, the function
+        declines execution.
+
+    Returns
+    -------
+    tuple
+        The k-tuple holding the IDs of the selected prototypes.
```
mortonjt: Same here, what are IDs?

phylogeny/prototypeSelection.py
```diff
@@ -0,0 +1,106 @@
...
+    -----
+    This is the reference implementation for an exact algorithm for the
+    prototype selection problem. It has an exponential runtime and will only
+    operate on small instances (< maxCombinationsToTest).
+    Idea: test all (n over k) combinations of selecting k elements from n with-
+          out replacement. Compute the objective for each such combination and
+          report the combination with maximal value.
+    '''
+    assert numPrototypes >= 2
+    assert numPrototypes < len(distanceMatrix.ids)
```
mortonjt: This is a little picky - is there any reason why the parenthesis are missing? Why not `assert(numPrototypes >= 2)`?

phylogeny/prototypeSelection.py
```diff
@@ -0,0 +1,106 @@
...
+    combinations = sp.special.binom(len(distanceMatrix.ids), numPrototypes)
+    if combinations >= maxCombinationsToTest:
+        raise Exception("Cowardly refuse to test %i combinations. Use a\
+                        heuristic implementation for instances with more than\
+                        %i combinations instead!"
+                        % (combinations, maxCombinationsToTest))
```
mortonjt: Here's something you might find interesting.

You can get rid of the backslashes as follows

``` python
        raise Exception( ("Cowardly refuse to test %i combinations. Use a "
                          "heuristic implementation for instances with more than "
                          "%i combinations instead!")
                          % (combinations, maxCombinationsToTest))
```

Placing strings into a tuple while leaving out the commas automatically concatenates the strings together.

phylogeny/tests/data/wrongFileformat.txt
```diff
@@ -0,0 +1,22 @@
+from unittest import TestCase, main
```
mortonjt: Can we ditch this?

phylogeny/tests/test_prototypeSelection.py
```diff
@@ -0,0 +1,139 @@
...
+
+class prototypeSelection(TestCase):
+    def setUp(self):
+        self.dm_nonNull = get_data_path('distMatrix_nonNull.txt')
+        self.dm_repIDs = get_data_path('distMatrix_repIDs.txt')
+        self.dm_asym = get_data_path('distMatrix_asym.txt')
+        # this file must not exists!
+        self.dm_noFile = get_data_path('noFile.txt')
+        # any file that is present, but not a DistanceMatrix
+        self.dm_wrongFormat = get_data_path('wrongFileformat.txt')
```
mortonjt: Scratch that, I see it here.

phylogeny/tests/test_prototypeSelection.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def test_exhaustive(self):
+        # check if execution is rejected if number of combination is too high
+        self.assertRaisesRegex(
+            Exception,
+            'Cowardly refuse to test ',
+            prototypeSelection_exhaustive,
+            self.dm20,
+            5,
+            maxCombinationsToTest=1000)
```
mortonjt: Hmm.  I have never seen a test run like this!

mortonjt:
Some minor comments.

.travis.yml
ekopylova: Phylonet is one of the phylogenetic tools we're testing, @mortonjt could you clarify what you mean? Thanks!

sjanssen2: Travis builds fail, because wget cannot retrieve the specified file since there are some issues with the according web certificate. The new flag --no-check-certificate makes wget ignore the certificate problem and enables the download of the *.jar. That might impose a security problem - but my bet is, that we can ignore it at the moment.

ekopylova: Yes, let's ignore it for the moment since PhyloNet will be conda installable soon.

phylogeny/prototypeSelection.py
sjanssen2: but they are totally ignored :-(
I would love to see an error if you provide an object of the wrong type to a function.

phylogeny/prototypeSelection.py
sjanssen2: shit, I don't know exactly what you are referring to.
Line 48: a Haskell habit. I like to check types, because it can save a lot of trouble, mainly because it clarifies expectations about the function.
Line 50: I could also return 0 as the sum of distances of only one point, but that would maybe confuse the user.
Line 51: the itertools.combinations operate on lists, i.e. duplicate elements are allowed. Thus, the sum for [A,B,C,A] would be larger than [A,B,C]

phylogeny/prototypeSelection.py
sjanssen2: I was confused by the documentation in skbio, which does not consistently use the two terms. I expect to work on a distance matrix (symmetric), otherwise my upcoming heuristics will fail. @ekopylova  I hope that our genomic distances will be symmetric.
I updated my documentation and always speak about "distance" matrices, now.

phylogeny/prototypeSelection.py
sjanssen2: Sorry for the confusion, it should read "iterable of str" as in skbio.
I mean the IDs of the elements for which the matrix holds distances. Do you think I should address them directly as "elements"?

phylogeny/prototypeSelection.py
sjanssen2: yes, changed my text to make that clear, btw. it must be < instead of <=

sjanssen2: fixed the typo

phylogeny/prototypeSelection.py
sjanssen2: changed to "The k-tuple holding selected prototypes, i.e. a sub-set of the elements in the distance matrix." to clarify

phylogeny/prototypeSelection.py
sjanssen2: no reason. But what is the more pythonic way? Avoid parenthesis like in if statements or add than like print() in python3 ?

phylogeny/prototypeSelection.py
sjanssen2: good to know, but it requires more keystrokes and I like the \ as indicator for a line break.

phylogeny/tests/test_prototypeSelection.py
sjanssen2: is that good or bad?

phylogeny/prototypeSelection.py
sjanssen2: Cool! But I have the feeling that the expensive part is filtering the distance matrix down to a matrix holding only the given elements. However, I now use np.tril(dm.filter(elements).data).sum() which is way clearer to read and it is ~2x faster.

phylogeny/prototypeSelection.py
sjanssen2: Line 51 removed: I am now using dm.filter() which checks for duplicates. 
Line 50 removed: using numpy.tril also works on a single element, but not on an empty list

wasade: What about py3 function annotations? They aren't enforced but puts
expectations on dev and can be checked later if necessary (ie do some ast
work). Qiime2 is doing a lot with the annotations

Also, isinstance is preferred over type. And note that that introspection
can be expensive in Python

sjanssen2: so, what style should I use here? Personally, I would like to annotate every type wherever possible. Is that sufficient and can I thus remove the (maybe expensive) type checking or should it remain, because python (at the moment) does not enforce types?

phylogeny/tests/data/wrongFileformat.txt
sjanssen2: no

phylogeny/tests/test_prototypeSelection.py
sjanssen2: exactly

phylogeny/prototypeSelection.py
wasade: Guido is working on compile time type checking but still in dev and not in
Python proper:

http://mypy-lang.org

Doing the annotations does hurt but is above and beyond what we do, for
instance, in scikit bio (although I could see a push to put them in since
that project is now py3 only).

It doesn't hurt. But I so recommend against explicit type checks except
perhaps in outer, user facing methods or high level API calls. It will
consume a large amount of runtime if they're done within a loop or nested
loop (generally aim to avoid those checks deep in the call stack)

On Sep 2, 2016 09:43, "Stefan Janssen" notifications@github.com wrote:

In phylogeny/prototypeSelection.py
https://github.com/biocore/WGS-HGT/pull/32#discussion_r77376519:

> -    elements: list or set
> -        list or set of element IDs for which the sum of distances is computed
> -    dm: skbio.stats.distance.DistanceMatrix
> -        pairwise dissimilarity matrix.
>   +
> -    Returns
> -    -------
> -    float:
> -        the sum of all pairwise distances of dm for IDs in elements
> -    '''
>   +
> -    # some assertions
> -    if type(dm) != DistanceMatrix:
> -        raise TypeError('dm is not of type "DistanceMatrix"')
> -    assert(len(elements) > 1)
> -    assert(len(elements) <= len(dm.ids))

so, what style should I use here? Personally, I would like to annotate
every type wherever possible. Is that sufficient and can I thus remove the
(maybe expensive) type checking or should it remain, because python (at the
moment) does not enforce types?

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
https://github.com/biocore/WGS-HGT/pull/32/files/f12a6fa86bd8fbfda95231d3b1a11d318a85014d#r77376519,
or mute the thread

sjanssen2: using mypy now. Awsome! We are coming close to Haskell types ;-)
Thus, I removed the explicit type check in my code and will relay on future python development for more generic type checks.

sjanssen2: arrrg. I just noticed that type annotation cause errors on the travis build for python 3.4
What to do now? Ignore that problem or remove the beautiful type annotations?

wasade: Only support 3.5? What's QIIME 2 doing?

phylogeny/prototypeSelection.py
```diff
@@ -0,0 +1,99 @@
...
+For completeness, the exact but exponential algorithm is implemented, too.
+  "prototypeSelection_exhaustive"
+
+[1] Gamez, J. Esteban, François Modave, and Olga Kosheleva.
+    "Selecting the most representative sample is NP-hard:
+     Need for expert (fuzzy) knowledge."
+    Fuzzy Systems, 2008. FUZZ-IEEE 2008.
+"""
+
+from skbio.stats.distance import DistanceMatrix
```
josenavas: Minor, but the numpy standards and/or PEP8 defines a specific order in which the imports should be done:

```
standard library imports 
blank line
3rd party tools imports
blank line
local project imports
```

Following this rule, your imports should look like:

``` python
from typing import Sequence, Tuple
import itertools  # can you use from itertools import XXX rather than importing the entire module?

import numpy as np
import scipy as sp
from skbio.stats.distance import DistanceMatrix
```

phylogeny/prototypeSelection.py
```diff
@@ -0,0 +1,99 @@
...
+import itertools
+import numpy as np
+import scipy as sp
+from typing import Sequence, Tuple
+
+
+def distanceSum(elements: Sequence[str], dm: DistanceMatrix) -> float:
```
josenavas: Minor but [PEP8](https://www.python.org/dev/peps/pep-0008/#function-names) defines variable and method names to use lowercase words separated by underscores, rather than mixedCase.
`distanceSum -> distance_sum`. Applies to all instances in the code (i.e. I will not comment this issue downstream)

phylogeny/prototypeSelection.py
```diff
@@ -0,0 +1,99 @@
...
+    Notes
+    -----
+    This is the reference implementation for an exact algorithm for the
+    prototype selection problem. It has an exponential runtime and will only
+    operate on small instances (< maxCombinationsToTest).
+    Idea: test all (n over k) combinations of selecting k elements from n with-
+          out replacement. Compute the objective for each such combination and
+          report the combination with maximal value.
+    '''
+    assert(numPrototypes >= 2)
```
josenavas: How often do you expect this asserts to be failing? Ideally you want to raise a specific exception with a meaningful messages, so other developers understand what is going on with the code (applies to all)

phylogeny/prototypeSelection.py
```diff
@@ -0,0 +1,99 @@
...
+    assert(numPrototypes < len(distanceMatrix.ids))
+    combinations = sp.special.binom(len(distanceMatrix.ids), numPrototypes)
+    if combinations >= maxCombinationsToTest:
+        raise Exception("Cowardly refuse to test %i combinations. Use a \
```
josenavas: Makes sense to have a "ValueError" rather than a generic Exception?

phylogeny/prototypeSelection.py
```diff
@@ -0,0 +1,99 @@
...
+
+    Notes
+    -----
+    This is the reference implementation for an exact algorithm for the
+    prototype selection problem. It has an exponential runtime and will only
+    operate on small instances (< maxCombinationsToTest).
+    Idea: test all (n over k) combinations of selecting k elements from n with-
+          out replacement. Compute the objective for each such combination and
+          report the combination with maximal value.
+    '''
```
josenavas, line 106: The documentation is missing a "Raises" section, in which the raised errors are explained

phylogeny/tests/test_prototypeSelection.py
```diff
@@ -0,0 +1,143 @@
...
+        self.assertAlmostEqual(66.94, distanceSum(res, self.dm20))
+
+        res = prototypeSelection_exhaustive(self.dm20, 19)
+        self.assertCountEqual(
+            ('A', 'B', 'C', 'D', 'E', 'F', 'G', 'I', 'J', 'K', 'L', 'M', 'N',
+             'O', 'P', 'Q', 'R', 'S', 'T'),
+            res)
+        self.assertAlmostEqual(74.1234, distanceSum(res, self.dm20))
+
+    def test_wellformedDistanceMatrix(self):
```
josenavas: @sjanssen2 looks like all this tests are just checking that DistanceMatrix.read is doing what is supposed to do? If so, I don't think that it needs to be tested here, since such function is already tested in scikit-bio. If you have the feeling that the tests in there are not exhaustive enough, you can submit a PR there to extend them!

phylogeny/tests/test_prototypeSelection.py
josenavas: As long as the tests are executed correctly, I think it is ok! Have you check a case in which the test fails?

josenavas:
Thanks @sjanssen2 !! Some comments! Also note that travis is currently failing.

phylogeny/tests/test_prototypeSelection.py
sjanssen2: implicitly by the testing correctness of the return value of the function in the following tests.

phylogeny/prototypeSelection.py
sjanssen2: it is now PEP8 compliant and I import only the one function I need from itertools. Good catch!

phylogeny/prototypeSelection.py
sjanssen2: a lot of work, but I changed all my names.

phylogeny/prototypeSelection.py
sjanssen2: True, but I do not really know what the best matching Exception in this case would be.

phylogeny/prototypeSelection.py
sjanssen2: true

phylogeny/tests/test_prototypeSelection.py
sjanssen2: You are right in principle, but should SKbio ever change this behaviour, I want to make sure I will see this. My function relay on those properties and would silently return wrong results (instead of raising errors). Thus, I'd like to keep those tests.

phylogeny/prototypeSelection.py
sjanssen2: added this part

sjanssen2:
Travis fails for python 3.4 because this version of the language does not work with type annotations. Since I really like them, we should discuss if we need python 3.4 compatibility here.

phylogeny/tests/test_prototypeSelection.py
josenavas: I understand your concerns. @wasade @antgonza @ElDeveloper
This is not what we usually do in our projects, but if you agree with this I'm happy to leave as it is.

wasade: Skbio has stated strong API constraints for how these things change. Not necessary. All public methods are decorated with stability information

wasade: Also, a distance matrix is symmetric...

phylogeny/prototypeSelection.py
sjanssen2: I decided to remove the type annotation from the signatures for now and add them as comments

phylogeny/tests/test_prototypeSelection.py
sjanssen2: Aren't unit tests always checking things that we take for granted and then they save our asses because we were wrong with those assumptions? I am a bit surprised that you argue for removing some tests. Even if we expect them to always pass, in what way do they hurt?

wasade: It would be more appropriate to push theses tests into scikit-bio as a problem there impacts many more users than a problem here; catching it at a lower level would be better. The test suite for this object is already [comprehensive](https://github.com/biocore/scikit-bio/blob/master/skbio/stats/distance/tests/test_base.py), although any corner cases not caught would be welcomed. 

Scikit-bio takes an aggressive approach to regression testing, so it would be a catastrophic failure on the review and build process for a release to come out with a bug along these. If one were to happen, it would almost certainly trigger an immediate bugfix release. 

Unit tests are pragmatic. There are an infinite number of tests that can be performed on even the simplest of programs but that does not mean they are all equally worth the developer effort to put in place and maintain.

wasade: I confess, I also really really enjoy deleting code. It's soooo satisfying...

sjanssen2: and soooo motivating for the guy that wrote the code ....
I removed those test, and should there ever be a problem with this object, I know whom to blame :-)

phylogeny/prototypeSelection.py
```diff
@@ -0,0 +1,125 @@
...
+# from typing import Sequence, Tuple
+from itertools import combinations
+
+import numpy as np
+import scipy as sp
+from skbio.stats.distance import DistanceMatrix
+
+
+def distance_sum(elements, dm):
+    # function signature with type annotation for future use with python >= 3.5
```
ekopylova: Could this comment be added to a ```Notes``` section?

phylogeny/prototypeSelection.py
```diff
@@ -0,0 +1,125 @@
...
+    float:
+        the sum of all pairwise distances of dm for IDs in elements
+    '''
+
+    return np.tril(dm.filter(elements).data).sum()
+
+
+def prototype_selection_exhaustive(dm, num_prototypes,
+                                   max_combinations_to_test=200000):
+    # function signature with type annotation for future use with python >= 3.5
```
ekopylova: same here, ```Notes```?

ekopylova:
@sjanssen2 is this ready to merge?

sjanssen2:
I think it is ready now. I just moved the function annotation comments into "Notes" sections as you suggested.

phylogeny/tests/test_prototypeSelection.py
wasade: :)

PR #34, opened 2016-09-07
added a heuristic function to select prototypes.

sjanssen2:
Thanks @ekopylova @mortonjt @josenavas @wasade for reviewing my first pull request for prototype selection. The methods were rather boring. 
But with this pull request I add a first heuristic function to select prototypes. I'd be happy to get another round of valuable feedback from you guys.

### DIFF
--- a/phylogeny/prototypeSelection.py
+++ b/phylogeny/prototypeSelection.py
@@ -9,9 +9,10 @@ Here, we define "best represents" as maximizing the sum of distances between
 all points, i.e. sum {over a,b in S} DM[a,b]. This is our objective function.
 
 This problem is known to be NP-hard [1], thus we need to resort to heuristics.
-In the future, this module will implement several heuristics, whose quality can
-be measured by the objective function for each problem instance, since there is
-no global winner.
+This module implements several heuristics, whose quality can be measured by
+the objective function for each problem instance, since there is no global
+winner. Currently implemented are:
+ - prototype_selection_constructive_maxdist
 For completeness, the exact but exponential algorithm is implemented, too.
   "prototype_selection_exhaustive"
 
@@ -37,14 +38,14 @@ def distance_sum(elements, dm):
     Parameters
     ----------
     elements: sequence of str
-        list or elements for which the sum of distances is computed
+        List or elements for which the sum of distances is computed.
     dm: skbio.stats.distance.DistanceMatrix
-        pairwise distance matrix.
+        Pairwise distance matrix.
 
     Returns
     -------
     float:
-        the sum of all pairwise distances of dm for IDs in elements
+        The sum of all pairwise distances of dm for IDs in elements.
 
     Notes
     -----
@@ -62,8 +63,7 @@ def prototype_selection_exhaustive(dm, num_prototypes,
     Parameters
     ----------
     dm: skbio.stats.distance.DistanceMatrix
-        pairwise distances for all elements in the full set S.
-        Must be symmetric and non-hollow.
+        Pairwise distances for all elements in the full set S.
     num_prototypes: int
         Number of prototypes to select for distance matrix.
         Must be >= 2, since a single prototype is useless.
@@ -75,7 +75,7 @@ def prototype_selection_exhaustive(dm, num_prototypes,
 
     Returns
     -------
-    sequence of str
+    list of str
         A sequence holding selected prototypes, i.e. a sub-set of the
         elements in the distance matrix.
 
@@ -102,7 +102,7 @@ def prototype_selection_exhaustive(dm, num_prototypes,
 
     function signature with type annotation for future use with python >= 3.5:
     def prototype_selection_exhaustive(dm: DistanceMatrix, num_prototypes: int,
-    max_combinations_to_test: int=200000) -> Sequence[str]:
+    max_combinations_to_test: int=200000) -> List[str]:
     '''
     if num_prototypes < 2:
         raise ValueError(("'num_prototypes' must be >= 2, since a single "
@@ -124,7 +124,7 @@ def prototype_selection_exhaustive(dm, num_prototypes,
         d = distance_sum(s, dm)
         if d > max_dist:
             max_dist, max_set = d, s
-    return max_set
+    return list(max_set)
 
 
 def prototype_selection_constructive_maxdist(dm, num_prototypes):
@@ -150,17 +150,12 @@ def prototype_selection_constructive_maxdist(dm, num_prototypes):
 
     Returns
     -------
-    sequence of str
+    list of str
         A sequence holding selected prototypes, i.e. a sub-set of the
         elements in the distance matrix.
 
     Raises
     ------
-    RuntimeError
-        Combinatorics explode even for small instances. To save the user from
-        waiting (almost) forever, this function declines execution if the
-        number of combinations to test are too high,
-        i.e. > max_combinations_to_test
     ValueError
         The number of prototypes to be found should be at least 2 and at most
         one element smaller than elements in the distance matrix. Otherwise, a
@@ -173,7 +168,7 @@ def prototype_selection_constructive_maxdist(dm, num_prototypes):
             where the dm holds 27,398 elements
     function signature with type annotation for future use with python >= 3.5:
     def prototype_selection_constructive_maxdist(dm: DistanceMatrix,
-    num_prototypes: int) -> Sequence[str]:
+    num_prototypes: int) -> List[str]:
     '''
     if num_prototypes < 2:
         raise ValueError(("'num_prototypes' must be >= 2, since a single "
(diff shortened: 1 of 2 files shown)


phylogeny/prototypeSelection.py
```diff
@@ -125,3 +125,82 @@ def prototype_selection_exhaustive(dm: DistanceMatrix, num_prototypes: int,
         if d > max_dist:
             max_dist, max_set = d, s
     return max_set
+
+
+def prototype_selection_constructive_maxdist(dm, num_prototypes, steps=10):
+    '''Heuristically select k prototypes for given distance matrix.
```
wasade, line 131: Can this be a one liner followed by more detail? eg

``` python
def foo(blah):
    """Some one line description

    More detail about all the things

    Parameters
    ----------
    blah : foo
        stuff and stuff

    Notes
    ------
    Some other note about something unexpected or whatever
    """
```

See multi-line docstrings in [PEP257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings) as laid out by the BDFL for the one-liner. The rest of the structure though is numpydoc

phylogeny/prototypeSelection.py
```diff
@@ -125,3 +125,82 @@ def prototype_selection_exhaustive(dm: DistanceMatrix, num_prototypes: int,
...
+       Greedily grow the set of prototypes by adding the element with the
+       largest sum of distances to the non-prototype elements.
+
+    Parameters
+    ----------
+    dm: skbio.stats.distance.DistanceMatrix
+        pairwise distances for all elements in the full set S.
```
wasade: pairwise -> Pairwise

phylogeny/prototypeSelection.py
```diff
@@ -125,3 +125,82 @@ def prototype_selection_exhaustive(dm: DistanceMatrix, num_prototypes: int,
...
+        Must be symmetric and non-hollow.
```
wasade: Isn't this implicit?

phylogeny/prototypeSelection.py
```diff
@@ -125,3 +125,82 @@ def prototype_selection_exhaustive(dm: DistanceMatrix, num_prototypes: int,
...
+    num_prototypes: int
+        Number of prototypes to select for distance matrix.
+        Must be >= 2, since a single prototype is useless.
+        Must be smaller than the number of elements in the distance matrix,
+        otherwise no reduction is necessary.
+
+    Returns
+    -------
+    sequence of str
```
wasade: it looks like this is list of str?

phylogeny/prototypeSelection.py
sjanssen2: corrected

phylogeny/prototypeSelection.py
sjanssen2: it is. I removed that line.

phylogeny/prototypeSelection.py
sjanssen2: I am not completely sure what a "sequence of " is. I thought it is a generic name for list, tuple, array-like. Since the element IDs of the distanceMatrix must be provided as a "sequence of str" I'd like to return this type.

phylogeny/prototypeSelection.py
sjanssen2: Didn't know that. Thanks for pointing me to it. I rearranged my documentation to adhere to this convention. Let me know if it suits now.

phylogeny/prototypeSelection.py
sjanssen2: casting to tuple now, to have the same type as "prototype_selection_exhaustive" and "distanceMatrix.ids", which is described as "sequence of str" in the skbio doc.

wasade: as input, it would be appropriate to describe as "Iterable of str" which suggests `list`, `tuple`, `generator`, etc would work. But in this case, this is describing the return type, which presumably is specific?

phylogeny/prototypeSelection.py
```diff
@@ -125,3 +125,82 @@ def prototype_selection_exhaustive(dm: DistanceMatrix, num_prototypes: int,
...
+    # being covered, i.e. prototypes
+    distance, resSet = dm.data.max(), list(np.unravel_index(dm.data.argmax(),
+                                           dm.data.shape))
+    uncovered[resSet] = np.False_
+
+    # repeat until enough prototypes have been selected:
+    #  the new prototype is the element that has maximal distance sum to all
+    #  non-prototype elements in the distance matrix.
+    while dm.shape[0] - np.sum(uncovered) < num_prototypes:
+        maxElmIdx = (dm.data[list(resSet), :].sum(axis=0) * uncovered).argmax()
```
wasade: I dont believe the call to `list` is necessary since it appears to be a list already?

phylogeny/prototypeSelection.py
```diff
@@ -125,3 +125,82 @@ def prototype_selection_exhaustive(dm: DistanceMatrix, num_prototypes: int,
...
+        uncovered[maxElmIdx] = np.False_
+        resSet.append(maxElmIdx)
+
+    # return the ids of the selected prototype elements
+    return [dm.ids[idx] for idx, x in enumerate(uncovered) if x == np.False_]
```
wasade: does `if not x` work?

wasade: if `dm.ids` is `np.array`, I believe this can be expressed as:

``` python
return dm.ids[np.argwhere(~uncovered)]
```

wasade:
Few more small comments :) CI is failing as well btw

sjanssen2:
Travis passes but the coverage decreases. I think it is not my fault, but in benchmark/tests/test_simulate_hgts.py

phylogeny/prototypeSelection.py
sjanssen2: dm.ids is a tuple of strings. Is it worth to first convert the tuple into an array here?

phylogeny/prototypeSelection.py
sjanssen2: ok, I changed it to a "list of str"

phylogeny/prototypeSelection.py
sjanssen2: you are right. Removed list here

josenavas:
@sjanssen2 the coveralls issue is not your fault - it looks like the project is missing the .coveragerc file so it is not checking correctly the coverage - I will do a PR soon fixing this so this doesn't block your development and I'm already familiar with such file :-D

sjanssen2:
oh perfect. Thanks @josenavas I really appreciate this!

josenavas:
Done: https://github.com/biocore/WGS-HGT/pull/35

phylogeny/prototypeSelection.py
```diff
@@ -125,3 +125,82 @@ def prototype_selection_exhaustive(dm: DistanceMatrix, num_prototypes: int,
...
+
+    Returns
+    -------
+    sequence of str
+        A sequence holding selected prototypes, i.e. a sub-set of the
+        elements in the distance matrix.
+
+    Raises
+    ------
+    RuntimeError
```
josenavas: @sjanssen2 I don't see where this error is raised in the function

phylogeny/prototypeSelection.py
josenavas: I think this change is not being reflected on github - maybe a missing "git push"?

phylogeny/prototypeSelection.py
sjanssen2: you are right, dump copy and paste mistake :-/

phylogeny/prototypeSelection.py
```diff
@@ -125,3 +125,82 @@ def prototype_selection_exhaustive(dm: DistanceMatrix, num_prototypes: int,
...
+                          " elements in the distance matrix, otherwise no "
+                          "reduction is necessary."))
+
+    # initially mark all elements as uncovered, i.e. as not being a prototype
+    uncovered = np.asarray([np.True_] * dm.shape[0])
+
+    # the first two prototypes are those elements that have the globally
+    # maximal distance in the distance matrix. Mark those two elements as
+    # being covered, i.e. prototypes
+    distance, resSet = dm.data.max(), list(np.unravel_index(dm.data.argmax(),
```
josenavas: minor but `resSet` -> `res_set`
I recognize that if you see my Haskell code you would not like it hahaha ;)

josenavas: oh! Also, it is more readable (and I think more pythonic) if written in two lines:

``` python
distance = dm.data.max()
res_set = list(np.unravel_index(dm.data.argmax(), dm.data.shape))
```

phylogeny/prototypeSelection.py
josenavas: `maxElmIdx` -> `max_elm_idx`

josenavas:
Just a couple of small comments - note that I'm not familiar with the functionality so I'm just checking style issues :-)

Also, I noted that this may slipped through in a previous PR, but the name of the file is `prototypeSelection.py`. I think it is more python compliant to be `prototype_selection.py`. However, don't bother on changing it right now - I think you can do a small PR later in which only that change is done :-)

phylogeny/prototypeSelection.py
josenavas: Not a problem! We all love ctrl-c, ctrl-v too much hahaha

phylogeny/prototypeSelection.py
wasade: oh, that's lame but makes sense I think (likely done for immutability). conversion would likely just introduce overhead so don't recommend it

sjanssen2:
Hey @josenavas @wasade, I commited some updates where I hopefully resolved all your concerns. Could you quickly check it out. I'd like to get this merged to be able to continue adding further methods. Thanks again for all your valuable comments!

phylogeny/prototypeSelection.py
```diff
@@ -124,4 +124,77 @@ def prototype_selection_exhaustive(dm: DistanceMatrix, num_prototypes: int,
...
+    # maximal distance in the distance matrix. Mark those two elements as
+    # being covered, i.e. prototypes
+    distance = dm.data.max()
+    res_set = list(np.unravel_index(dm.data.argmax(), dm.data.shape))
+    uncovered[res_set] = np.False_
+
+    # repeat until enough prototypes have been selected:
+    #  the new prototype is the element that has maximal distance sum to all
+    #  non-prototype elements in the distance matrix.
+    while dm.shape[0] - np.sum(uncovered) < num_prototypes:
```
wasade: minor, but a counter would probably be more efficient than `sum` as the latter requires walking an array on each iteration of the loop

PR #36, opened 2016-09-13
Prototype selection constructive protoclass

sjanssen2:
Another heuristic approach to solve the prototype selection problem.


sjanssen2:
@wasade @josenavas ready for a new heuristic?

sjanssen2:
@wasade @josenavas I need some reviews please!

phylogeny/prototypeSelection.py
```diff
@@ -201,3 +202,173 @@ def prototype_selection_constructive_maxdist(dm: DistanceMatrix,
...
+    prototypes = []
+    for i in range(steps):
+        oldDirection = direction
+        oldEpsilon = epsilon
+
+        # increase the stepsize in each iteration to converge faster
+        stepSize *= 1.1
+        # call the protoclass with a defined epsilon
+        prototypes = _protoclass(dm, epsilon)
+        # check if direction of epsilon changes has changed
```
qiyunzhu, line 353: Will it be possible that `len(prototypes) == num_prototypes`? In this case `direction` will be 0. Does it matter?

phylogeny/prototypeSelection.py
```diff
@@ -201,3 +202,173 @@ def prototype_selection_constructive_maxdist(dm: DistanceMatrix,
...
+    # (row)
+    B = dm.data < epsilon
+    # tracks which elements are covered by prototypes
+    covered = [0] * dm.shape[0]
+    # score is the number of other elements that falls within the epsilon ball
+    scores = B.sum(axis=0)
+    # found prototypes
+    prototypes = []
+
+    i = 0
```
qiyunzhu: This `i` isn't used anywhere in the while loop.

phylogeny/prototypeSelection.py
```diff
@@ -201,3 +202,173 @@ def prototype_selection_constructive_maxdist(dm: DistanceMatrix,
...
+    while(True):
```
qiyunzhu: I wonder if `while True:` is the more Pythonic way, but I am not sure... Anyway this is trivial.

qiyunzhu (commented):
@sjanssen2 explained to me the mathematical background of this algorithm and his actual implementation of it. Sounds very reasonable to me. The code looks clean and understandable, with sufficient documentation. I have only trivial comments.

phylogeny/prototypeSelection.py
```diff
@@ -201,3 +202,173 @@ def prototype_selection_constructive_maxdist(dm: DistanceMatrix,
...
+    [1] Jacob Bien and Robert Tibshirani.
+        "Prototype selection for interpretable classification."
+        The Annals of Applied Statistics (2011): 2403-2424.
+    '''
+
+    # is an element (column) covered by the epsilon ball of another element
+    # (row)
+    B = dm.data < epsilon
+    # tracks which elements are covered by prototypes
+    covered = [0] * dm.shape[0]
```
wasade: `np.zeros(dm.shape[0], type=bool)` ?

phylogeny/prototypeSelection.py
```diff
@@ -201,3 +202,173 @@ def prototype_selection_constructive_maxdist(dm: DistanceMatrix,
...
+        i += 1
+
+        # candidate for a new prototype is the element whose epsilon ball
+        # covers most other elements.
+        idx_max = scores.argmax()
+        if (scores[idx_max] > 0):
+            # candidate is new prototype, add it to the list
+            prototypes.append(idx_max)
+            # which elements have been just covered by the new prototype
+            justcovered = B[:, idx_max] & np.logical_not(covered)
```
wasade, line 260: I _think_ doing `np.logical_and(B[:, idx_max], np.logical_not(covered))`, although it could be the case that numpy is appropriately handling the operator too...

phylogeny/prototypeSelection.py
```diff
@@ -201,3 +202,173 @@ def prototype_selection_constructive_maxdist(dm: DistanceMatrix,
...
+            # update the global list of ever covered elements
+            covered += justcovered
```
wasade, line 262: should this be `covered = np.logical_or(covered, justcovered)`?

phylogeny/prototypeSelection.py
sjanssen2: ups, this is a left over from testing :-/

phylogeny/prototypeSelection.py
sjanssen2: I guess you are right here. I removed the parenthesis.

qiyunzhu:
@sjanssen2 would you like to take a look at the comments and respond? Thanks!

phylogeny/prototypeSelection.py
sjanssen2: The whole method is for finding an epsilon value for which len(prototypes) == num_prototypes. Thus it does matter and indicates the successful end of the search. See a few lines below.

phylogeny/prototypeSelection.py
sjanssen2: I changed as you recommended, but what is the advantage?

phylogeny/prototypeSelection.py
sjanssen2: I tested and saw that both versions work and produce the same results. I prefer my line, because it is shorter.

phylogeny/prototypeSelection.py
sjanssen2: same as before. Both work, I think my shorter lines is more readable.

phylogeny/prototypeSelection.py
qiyunzhu: I like @sjanssen2 's original solution, because it is a more general solution. One can do `['whatever'] * n`.

qiyunzhu (approved):

qiyunzhu:
oops! @sjanssen2 looks like your code is 100% covered. Then how did this problem occur? @ekopylova any thoughts?

qiyunzhu:
Cool! @wasade how do @sjanssen2 's responses sound to you? Thanks!

wasade:
:+1:

PR #40, opened 2016-10-17
simulate_hgts.py adds GenBank output

qiyunzhu:
Now the script can generate GenBank output. The recipient GenBank file contains all original metadata (with sequence size updated). The loci information was regenerated from the recipient gene list. They read like:

```
         gene        39119..40364
                     /locus_tag=WP_020915760.1
          CDS        39119..40364
                     /locus_tag=WP_020915760.1
                     /protein_id=WP_020915760.1
                     /translation=MISHNKVLFFKKFKLNTFIKKFLNKNFF...
```

Is this the expected pattern?
Also, please double-check the changes I made in generating the recipient genome (line # 278-280, 376-378). I am not sure if that's correct.
Thanks!

### DIFF
--- a/benchmark/simulate_hgts.py
+++ b/benchmark/simulate_hgts.py
@@ -96,7 +96,7 @@ def launch_orthofinder(proteomes_dir, threads, verbose=False):
     verbose: boolean
         if True, run in verbose mode
     """
-    if verbose: 
+    if verbose:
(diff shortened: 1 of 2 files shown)


ekopylova:
@antgonza would you have a chance to review this?

qiyunzhu:
@antgonza Thanks!

PR #41, opened 2016-10-19
Merged reformat_input_gi into reformat_input

qiyunzhu:
As the title indicates, also added GeneMark reformat function (it is a subset of the EGID reformat function). Note that because GeneMark and EGID takes GenBank file instead of trees, the "required" properties of the click options were all set to False.


qiyunzhu:
Update: Never mind. Solved with a trick :)

qiyunzhu:
Hi all, anyone would like to review (and maybe merge) it? The next one is coming :)

benchmark/reformat_input.py
```diff
@@ -337,19 +339,192 @@ def reformat_treepuzzle(gene_tree,
...
+    Returns
+    -------
+    tuple of (
+        skbio.Sequence,
+            Genome sequence and metadata
+        dict of { list of [ string, int, int, string ] }
+            Gene name : translation, start, end, and strand
+    )
+    """
+    loci = []
```
antgonza, line 359: I think lines 359->371 can be merged. In the current code you open the file, count the number of sequences, close it, and then ask each of the sequences to Sequence.read. I think you could simply do: `for i, gb  in enumerate(Sequence.read(genbank_fp, format='genbank'))`. Obviously you could not use i, if not actually need it.

benchmark/reformat_input.py
```diff
@@ -337,19 +339,192 @@ def reformat_treepuzzle(gene_tree,
...
+    output_dir: string
+        output directory path
+
+    Notes
+    -----
+    The reason this script does not use skbio's GenBank write function, is
+    that EGID cannot parse skbio-written GenBank format. Perhaps the reason
+    is the differential pattern of leading white spaces in the feature table.
+    """
+    output_f = open(join(output_dir, 'id.gbk'), 'w')
```
antgonza, line 415: Are we worry about file overwriting (AKA using the same filename more than once)?

benchmark/reformat_input.py
```diff
@@ -337,19 +339,192 @@ def reformat_treepuzzle(gene_tree,
...
+    tmp_fp = join(output_dir, 'id.tmp')
+    DNA.write(gb, tmp_fp, format='genbank')
```
antgonza, line 417: Here you write to tmp_fp the file and the you open it and rewrite it to output_f, right? Is that the expected behavior?

benchmark/reformat_input.py
```diff
@@ -337,19 +339,192 @@ def reformat_treepuzzle(gene_tree,
...
+    with open(tmp_fp, 'r') as input_f:
+        output_f.write(input_f.readline())
+    output_f.write('FEATURES             Location/Qualifiers\n')
+    gid = 1
+    for (gene, l) in sorted(genes.items(), key=lambda x: x[1][1]):
```
antgonza, line 422: Then here you write at the end of the file all the FEATURES, etc. I think there is something odd going on here.

benchmark/reformat_input.py
```diff
@@ -337,19 +339,192 @@ def reformat_treepuzzle(gene_tree,
...
+        gid += 1
+    with open(tmp_fp, 'r') as input_f:
+        next(input_f)
+        for line in input_f:
+            output_f.write(line)
+    output_f.close()
+    remove(tmp_fp)
+
+
+def reformat_egid(genbank_fp,
```
antgonza, line 444: I think this method suffers from the same errors than the previous ones, could you also review this and the following methods?

benchmark/reformat_input.py
qiyunzhu: I once did the way you (and another guy) recommended. However, I noticed a significant performance loss. On July 17 I wrote the following statement in an email to Jenya: _I realized that for a multi-sequence GenBank file, scikit-bio cannot parse all sequences together, but has to do this one per time (the developer can specify a sequence index). In my script I followed this mechanism to loop through all sequences. However, I found that the processing speed is substantially slower than splitting the master genbank files into multiple child files with one sequence each, and feed scikit-bio with them sequentially. For example for a 9-sequence file, the former took 18 seconds to parse while the later only took 4 seconds._

benchmark/reformat_input.py
qiyunzhu: Thanks for noting this. Yes it sounds odd. The logic is: scikit-bio can generate a GenBank file, but its format is not exactly compatible with EGID. Specifically, this incompatibility is in the feature table. Therefore, my solution was:
1) Use scikit-bio to write a GenBank structure without feature table to a temporary file.
2) Open this temp file, read the first line (locus table), and write into the output GenBank file.
3) Use Python to write the feature table into the output GenBank file.
4) Open the temp file again, read the rest lines (genome sequence), and write into the output GenBank file.
Does this sound a reasonable plan?

benchmark/reformat_input.py
qiyunzhu: See above.

benchmark/reformat_input.py
qiyunzhu: See above.

benchmark/reformat_input.py
qiyunzhu: That may be the point. If both EGID and GeneMark benchmark are executed simultaneously using the same output_dir, I am afraid that there may be a file name conflict. I am not too sure about the pipeline's design. @ekopylova do you have comments on this?

qiyunzhu (commented):
@antgonza Thank you!! You looked very carefully and found the logic oddness. I wrote these codes long before joining the KL. That was my first lesson.

benchmark/reformat_input.py
ekopylova: @RNAer @mortonjt do you know why there's such a performance loss using Antonio's suggestion vs. Qiyun's current implementation?

benchmark/reformat_input.py
```diff
@@ -337,19 +339,192 @@ def reformat_treepuzzle(gene_tree,
...
+    output_dir: string
+        output directory path
+
+    Notes
+    -----
+    GeneMark's acceptable input file format is FASTA (genome sequence).
+    """
+    (gb, genes) = _merge_genbank_seqs(genbank_fp)
+    DNA.write(gb, join(output_dir, 'id.fna'), format='fasta')
+    _write_genebank_file(gb, genes, output_dir)
```
ekopylova, line 508: @qiyunzhu could you remind me again what the difference is between the GenBank file you read and the one you write for GeneMark? I can understand reading in order to generate a nucleotide complete genome sequence (although a nucleotide complete genome sequence was initially required input to the workflow), however what is the purpose of writing if you already have a GenBank file? How is the new GenBank file different from what you've read?

benchmark/reformat_input.py
```diff
@@ -358,34 +533,46 @@ def reformat_treepuzzle(gene_tree,
...
 @click.option('--method', required=True,
               type=click.Choice(['trex', 'ranger-dtl',
                                  'riata-hgt', 'consel',
-                                 'darkhorse', 'wn-svm',
-                                 'genemark', 'hgtector',
+                                 'darkhorse', 'hgtector',
+                                 'genemark', 'egid',
                                  'distance-method', 'jane4',
-                                 'tree-puzzle']),
+                                 'wn-svm', 'tree-puzzle']),
```
ekopylova, line 545: could we remove wn-svm from this list?

benchmark/reformat_input.py
ekopylova: Can we add a check to `_main()` to verify `--output-dir` doesn't already exist and throw an error if it does?

benchmark/reformat_input.py
ekopylova: Do we know if it's only EGID that has problems reading skbio generated GenBank files? @RNAer @mortonjt have you seen this problem before? Currently we're using `scikit-bio=0.4.2dev` from @mortonjt's channel. Is there a later version we can try?

ekopylova (commented):
Overall looks good but we need to resolve the issue with skbio generated GenBank file. Specifically, if it's an issue with skbio's implementation it would be good know. @qiyunzhu what exact error are you seeing when passing this file to EGID? Perhaps @RNAer or  @mortonjt could give us more insight?

ekopylova:
Both Colombo and Pai-IDA from EGID use GenBank input.

Both tools work with the example NC_009504.gbk file provided with EGID, however I'm receiving errors when trying other genomes (e.g., NC_010473.gb Escherichia coli str. K12 substr. DH10B, complete genome from NCBI). Are we sure that the GenBank reading error is from skbio or there's something else going on?

qiyunzhu:
Hi @ekopylova , the difference in the GenBank format is:
Official gbk format (also the one generated by my homebrew function in reformat_input.py):

```
     gene            1..1317
                     /locus_tag=gene1
     CDS             1..1317
                     /locus_tag=gene1
                     /protein_id=WP_011672224.1
                     /translation=MNTIFSRITPLGNGTLCVIRISGKNVKFLIQKIVKKNIKEKIATFSKLFLDKECVDYAMIIFFKKPNTFTGEDIIEFHIHNNETIVKKIINYLLLNKARFAKAGEFLERRYLNGKISLIECELINNKILYDNENMFQLTKNSEKKIFLCIIKNLKFKINSLIICIEIANFNFSFFFFNDFLFIKYTFKKLLKLLKILIDKITVINYLKKNFTIMILGRRNVGKSTLFNKICAQYDSIVTNIPGTTKNIISKKIKILSKKIKMMDTAGLKIRTKNLIEKIGIIKNINKIYQGNLILYMIDKFNIKNIFFNIPIDFIDKIKLNELIILVNKSDILGKEEGVFKIKNILIILISSKNGTFIKNLKCFINKIVDNKDFSKNNYSDVKILFNKFSFFYKEFSCNYDLVLSKLIDFQKNIFKLTGNFTNKKIINSCFRNFCIGK
```

The format generated by scikit-bio 0.4.2 dev:

```
         gene        1..1317
                     /locus_tag=gene1
          CDS        1..1317
                     /locus_tag=gene1
                     /protein_id=WP_011672224.1
                     /translation=MNTIFSRITPLGNGTLCVIRISGKNVKFLIQKIVKKNIKEKIATFSKLFLDKECVDYAMIIFFKKPNTFTGEDIIEFHIHNNETIVKKIINYLLLNKARFAKAGEFLERRYLNGKISLIECELINNKILYDNENMFQLTKNSEKKIFLCIIKNLKFKINSLIICIEIANFNFSFFFFNDFLFIKYTFKKLLKLLKILIDKITVINYLKKNFTIMILGRRNVGKSTLFNKICAQYDSIVTNIPGTTKNIISKKIKILSKKIKMMDTAGLKIRTKNLIEKIGIIKNINKIYQGNLILYMIDKFNIKNIFFNIPIDFIDKIKLNELIILVNKSDILGKEEGVFKIKNILIILISSKNGTFIKNLKCFINKIVDNKDFSKNNYSDVKILFNKFSFFYKEFSCNYDLVLSKLIDFQKNIFKLTGNFTNKKIINSCFRNFCIGK
```

As you can see, the difference is the pattern of leading white spaces.
I also sent an Email to you, including the actual files and the Python codes.

ekopylova:
@qiyunzhu I see. @RNAer is the # of white spaces mentioned in the GenBank format rules?

RNAer:
I don't think the number of spaces is strict. if i remember correctly, there is no specification on this acoording to genbank format.

qiyunzhu:
@RNAer It may be true, but EGID complained. Therefore I did an ugly way...

ekopylova:
Yes, but as far as I understand, both programs in EGID (Colombo and Pai-IDA) have issues with this format. I'm wondering if this is something common and if the spaces matter in other tools as well.

ekopylova:
Pai-IDA uses Bioperl (Bio::SeqIO) to read GenBank files and throws the Warning "Unexpected error in feature table for  Skipping feature, attempting to recover" when reading a GenBank file output by skbio (I can provide this file if needed).

qiyunzhu:
@ekopylova That's a very careful observation! So the conflict is not between scikit-bio and EGID (a very specific application), but Bioperl (used by many people). This may be of concern.

benchmark/reformat_input.py
qiyunzhu: Sure!

antgonza:
Suggest opening an issue in Bioperl to let them know. In the meanwhile and only using one format, if possible, that works in both libraries ...

ekopylova:
Both Bioperl (from Pai-Ida) and Colombo

ekopylova:
Looks good otherwise, @qiyunzhu could you create an issue on WGS-HGT to replace the code once other libraries have been updated to resolve the issue?

RNAer:
are you guys able to pin down what causes that warning specifically? I am
not sure giving me the file will be helpful...

On Fri, Oct 21, 2016 at 2:33 PM, Evguenia Kopylova <notifications@github.com

RNAer:
how many spaces of indent they use? I am refactoring the genbank parser in
skbio and can make this change as well. is the indentation the only
difference here?

On Fri, Oct 21, 2016 at 2:21 PM, Evguenia Kopylova <notifications@github.com

qiyunzhu:
Hi @RNAer I just sent the files to you via Email. As I noticed, the only difference is the indentation, but my observation may not be thorough.

PR #43, opened 2016-10-25
Merge parse_output and added parse Genemark

qiyunzhu:


qiyunzhu:
I proceeded and did the similar thing to the parse_output script. Now it handles EGID, GeneMark as well as other tools. Note: for these two programs, an additional parameter: --genebank-fp, is necessary.

benchmark/parse_output.py
```diff
@@ -141,14 +142,109 @@ def parse_hgtector(input_f):
...
+                protein_id = feature['protein_id'].replace('\"', '')
+                if protein_id not in genes:
+                    loc = gb.interval_metadata.features[feature]
+                    start = loc[0][0] + 1
+                    end = loc[0][1]
+                    genes[protein_id] = (start, end)
+    genes_in_gi = {}
+    for line in input_f:
+        l = line.strip().split()
+        if len(l) < 2:
```
ekopylova, line 182: what does it mean if `len(l) < 2`, is that the header?

benchmark/parse_output.py
```diff
@@ -141,14 +142,109 @@ def parse_hgtector(input_f):
...
+            continue
+        (start, end) = (int(l[0]), int(l[1]))
+        for (gene, l) in genes.items():
+            if (l[0] >= start and l[1] <= end):
```
ekopylova: are `l[0]` and `l[1]` considered `int` here or should they be casted prior comparison to `start` and `end`?

benchmark/parse_output.py
```diff
@@ -141,14 +142,109 @@ def parse_hgtector(input_f):
...
+                    loc = gb.interval_metadata.features[feature]
+                    start = loc[0][0] + 1
+                    end = loc[0][1]
+                    genes[protein_id] = (start, end)
+    genes_in_gi = {}
+    for line in input_f:
+        l = line.strip().split()
+        if len(l) < 2:
+            continue
+        (start, end) = (int(l[0]), int(l[1]))
```
ekopylova: where do you use this tuple afterward?

benchmark/parse_output.py
```diff
@@ -141,14 +142,109 @@ def parse_hgtector(input_f):
...
+        l = line.strip().split()
+        if len(l) == 2 and l == ['#', 'Length']:
+            reading = True
+        elif reading and len(l) == 6 and l[5] == '2':
+            (start, end, strand) = (int(l[2].lstrip('<>')),
+                                    int(l[3].lstrip('<>')),
+                                    l[1])
+            for (gene, l) in genes.items():
+                if l[0] == start and l[1] == end and l[2] == strand:
+                    atypical_genes.append(gene)
```
ekopylova, line 239: should you also check what class the gene is (1 or 2) prior to appending it to `atypical_genes`?

ekopylova (commented):
@qiyunzhu a few small comments, thanks!

benchmark/parse_output.py
qiyunzhu: A valid GI definition line should at least contain start and end (two columns). Although an EGID output file never has anything other than GI definition lines, but in case there is an exception (e.g., an empty line), this code will make the process smooth.

benchmark/parse_output.py
qiyunzhu: Two lines below :)

benchmark/parse_output.py
qiyunzhu: I think they are `int` already. See lines 174-175.

benchmark/parse_output.py
qiyunzhu: The check was done in line 229 (`and l[5] == '2'`).

qiyunzhu (commented):
@ekopylova Thanks for review! I answered your questions. Meanwhile, I created a new PR that contains the GenBank update. Either one can be merged first.

qiyunzhu:
Hi @antgonza and @sjanssen2 will either of you be interested in taking a look at this PR? Thank you!

benchmark/parse_output.py
ekopylova: ah yes, I see my confusion. At line # 179 you used `l` to store a list and at line # 183 you're resetting `l` to store the `(start,end)` tuple. Does this affect your expected results anywhere?

benchmark/parse_output.py
ekopylova: you're right, do you mind adding a comment above that line to specify that? Thanks :)

qiyunzhu: No problem! I will add it :)

benchmark/parse_output.py
qiyunzhu: I guess my coding habit needs to be improved. Let me replace one `l` with something else to avoid confusion.

qiyunzhu:
@ekopylova Thanks! I just edited.

benchmark/parse_output.py
```diff
@@ -141,14 +142,110 @@ def parse_hgtector(input_f):
...
+    """
+    genes = {}
+    gb = Sequence.read(genbank_fp, format='genbank')
+    for feature in gb.interval_metadata.features:
+        if feature['type_'] == 'CDS':
+            if 'protein_id' in feature:
+                protein_id = feature['protein_id'].replace('\"', '')
+                if protein_id not in genes:
+                    loc = gb.interval_metadata.features[feature]
+                    start = loc[0][0] + 1
```
antgonza, line 175: Just out of curiosity, why + 1, and is this going to affect other tools/etc?

benchmark/parse_output.py
antgonza: Could you add a comment above that line about this? Perhaps something like this:
`# a valid GI definition should have at least 2 columns`

benchmark/parse_output.py
antgonza: I think the confusion comes because you are note actually creating a tuple, you are simply summarizing 2 lines of code into 1. Perhaps splitting is a good idea, obviously fine if not.

antgonza (commented):
A few comments.

benchmark/parse_output.py
qiyunzhu: In scikit-bio, this number is the actual start location - 1. I am not sure why scikit-bio is coded like this (perhaps because computer languages count from 0 instead of 1). This won't affect other tools.

benchmark/parse_output.py
qiyunzhu: Good! Will do.

benchmark/parse_output.py
qiyunzhu: You are right. Let me change then.

qiyunzhu (commented):
@antgonza Thank you!!!

benchmark/parse_output.py
antgonza: Can you add a comment? Just to be clear for the next dev ...

qiyunzhu: Sure! Done.

qiyunzhu:
@ekopylova and @antgonza would you like to check if all your comments have been addressed and the PR is safe to be merged? Thank you!

PR #45, opened 2016-11-04
edited Bash scripts for EGID and GeneMark

qiyunzhu:
I worked on the Bash scripts a little bit. Thanks for your attention! @ekopylova @antgonza @sjanssen2 @wasade

### DIFF
--- a/benchmark/launch_software.sh
+++ b/benchmark/launch_software.sh
@@ -25,7 +25,7 @@ working_dir=$(readlink -m $1)
 scripts_dir=$(readlink -m $2)
 # species tree in Newick format
 species_tree_fp=$3
-# reference genome in FASTA format (for compositional methods)
+# species genome in GenBank format (for compositional methods)
 species_genome_fp=$4
 # species HMM model (produced by GeneMarkS)
 species_model_fp=$5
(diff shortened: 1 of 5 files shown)


benchmark/launch_software.sh
```diff
@@ -77,6 +77,8 @@ gi_to_taxid_fp=${26}
 parse_hgts=${27}
 # EGID install directory
 egid_install_dp=${28}
+# GeneMark install directory
+genemark_install_dp=${29}
 # qsub params
 qsub="-q route -m abe -M jenya.kopylov@gmail.com -l nodes=1:ppn=${threads} -l walltime=230:00:00 -l pmem=10gb -l mem=20gb"
```
ekopylova, line 83: can we pass this as an argument as well?

ekopylova (commented):
one comment, otherwise looks good! Thanks @qiyunzhu

benchmark/launch_software.sh
qiyunzhu: Good idea! Actually as I read thru the whole set of Bash scripts, I thought that there are many things we need to tune up. This is only one of them. I am planning to do it starting from next PR. At the moment I am trying to deploy WGS-HGT in my Barnacle to see how a user would run it.

qiyunzhu:
@ekopylova Thanks for reviewing!

benchmark/launch_software.sh
qiyunzhu: Moreover, because there are so many arguments (>30), how about putting them into a configuration file, instead of command-line arguments? This not only saves trouble for the user, but also allow us not to pay attention to the order of arguments.

antgonza:
Looks good to me but now sure if @ekopylova question is actually answered.

qiyunzhu:
@antgonza Thanks! Let's see how Jenya thinks about it. @ekopylova

qiyunzhu:
@ekopylova Thanks!!

PR #46, opened 2016-11-08
added another heuristic for prototype selection, based on the p-media…

sjanssen2:
…n idea.

### DIFF
--- a/phylogeny/prototypeSelection.py
+++ b/phylogeny/prototypeSelection.py
@@ -435,7 +435,7 @@ def prototype_selection_constructive_pMedian(dm, num_prototypes):
     # the first prototype.
     prototypes.append(np.argmin(dm.data.sum(axis=1)))
 
-    # repeat adding prototypes until the desired number is found
+    # repeat adding prototypes until the desired number is found.
     while len(prototypes) < num_prototypes:
         # for each element, we compute the smallest distance sum to each
         # previously found prototype ...
--- a/phylogeny/prototypeSelection.py
+++ b/phylogeny/prototypeSelection.py
@@ -435,7 +435,7 @@ def prototype_selection_constructive_pMedian(dm, num_prototypes):
     # the first prototype.
     prototypes.append(np.argmin(dm.data.sum(axis=1)))
 
-    # repreat adding prototypes until the desired number is found
+    # repeat adding prototypes until the desired number is found
     while len(prototypes) < num_prototypes:
         # for each element, we compute the smallest distance sum to each
         # previously found prototype ...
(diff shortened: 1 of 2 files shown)


phylogeny/prototypeSelection.py
```diff
@@ -369,3 +370,81 @@ def prototype_selection_constructive_protoclass(dm: DistanceMatrix,
...
+        "Non-hierarchical clustering with MASLOC"
+        Pattern Recognition, 1983, Vol. 16, No. 5, pp. 507-516
+    '''
+    if num_prototypes < 2:
+        raise ValueError(("'num_prototypes' must be >= 2, since a single "
+                          "prototype is useless."))
+    if num_prototypes >= len(dm.ids):
+        raise ValueError(("'num_prototypes' must be smaller than the number of"
+                          " elements in the distance matrix, otherwise no "
+                          "reduction is necessary."))
```
qiyunzhu, line 429: Just curious, is there a mechanism to ensure that len(dm.ids) is always > 2?

phylogeny/prototypeSelection.py
```diff
@@ -369,3 +370,81 @@ def prototype_selection_constructive_protoclass(dm: DistanceMatrix,
...
+
+    # start with an empty list of prototypes
+    prototypes = []
+
+    # add the one element whose distance is smallest to all other elements as
+    # the first prototype.
+    prototypes.append(np.argmin(dm.data.sum(axis=1)))
+
+    # repreat adding prototypes until the desired number is found
```
qiyunzhu: "repreat" is a typo?

qiyunzhu (commented):
The structure of this new function pretty much resembles exisiting functions (e.g., prototype_selection_constructive_maxdist). Documentation is very clear. I just have a small question. Everything else looks good.

phylogeny/prototypeSelection.py
sjanssen2: Good point. I did not explicitly mention the case that the distance matrix itself hold only 1 or 2 elements. But it is captured by the existing checks. However the message might not point explicitly in that direction. Do you think I should add another exception or can we life with that edge case and an indirect report to the user of my functions?

phylogeny/prototypeSelection.py
sjanssen2: sure

phylogeny/prototypeSelection.py
qiyunzhu: Where can I find the existing check? If it already excludes the chance of len(dm.ids) <= 2 then I guess we don't need to add exception here?

sjanssen2: lines 423: if num_prototypes < 2: in combination with line 426 if num_prototypes >= len(dm.ids):

qiyunzhu: Oh I see your rationale. My bad. Your logic is quite solid.

qiyunzhu (commented):
Good

qiyunzhu:
@antgonza @wasade @ekopylova Would any of you like to review @sjanssen2 's code? Thanks!

qiyunzhu:
@wasade Thank you!

PR #47, opened 2016-11-12
Prototype selection destructive maxdist

sjanssen2:
Most promising heuristic for prototype selection

### DIFF
--- a/phylogeny/prototypeSelection.py
+++ b/phylogeny/prototypeSelection.py
@@ -13,6 +13,7 @@ This module implements several heuristics, whose quality can be measured by
 the objective function for each problem instance, since there is no global
 winner. Currently implemented are:
  - prototype_selection_constructive_maxdist
+ - prototype_selection_destructive_maxdist
  - prototype_selection_constructive_protoclass
  - prototype_selection_constructive_pMedian
 For completeness, the exact but exponential algorithm is implemented, too.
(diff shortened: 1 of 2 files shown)


phylogeny/prototypeSelection.py
```diff
@@ -369,3 +371,161 @@ def prototype_selection_constructive_protoclass(dm: DistanceMatrix,
...
+    currDists[minElmIdx], numRemain = np.infty, numRemain-1
+
+    # continue until only num_prototype elements are left
+    while (numRemain > num_prototypes):
+        # substract the distance to the removed element for all remaining
+        # elements
+        currDists -= dm.data[minElmIdx]
+        # find the next element to be removed, again as the one that is
+        # clostest to all others
+        minElmIdx = currDists.argmin()
```
qiyunzhu, line 524: "clostest" is a typo...

qiyunzhu (commented):
Good!

qiyunzhu:
This one will need some review too @wasade @antgonza @ekopylova if you would like.

wasade:
@sjanssen2 would it be possible to pull from upstream master? I think commits in here overlap with #46

phylogeny/prototypeSelection.py
sjanssen2: fixed. Thanks!

sjanssen2:
Merged with biocore master and fixed one typo.

qiyunzhu:
@sjanssen2 Thanks for merging! @wasade

phylogeny/prototypeSelection.py
```diff
@@ -448,3 +449,83 @@ def prototype_selection_constructive_protoclass(dm: DistanceMatrix,
         prototypes.append(np.asarray(minVals).argmin())
 
     return [dm.ids[idx] for idx in prototypes]
+
+
+def prototype_selection_destructive_maxdist(dm, num_prototypes):
+    '''Heuristically select k prototypes for given distance matrix.
+
+       Prototype selection is NP-hard. This is an implementation of a greedy
+       correctness heuristic: Start with the complete set and interatively
```
wasade: interatively -> iteratively

phylogeny/prototypeSelection.py
```diff
@@ -448,3 +449,83 @@ def prototype_selection_constructive_protoclass(dm: DistanceMatrix,
...
+    currDists = dm.data.sum(axis=1)
+
+    # the element to remove first is the one that has smallest distance to all
+    # other. "Removing" works by tagging its distance-sum as infinity. Plus, we
+    # decrease the number of available elements by one.
+    minElmIdx = currDists.argmin()
+    currDists[minElmIdx], numRemain = np.infty, numRemain-1
+
+    # continue until only num_prototype elements are left
+    while (numRemain > num_prototypes):
```
wasade, line 518: It seems like this block of code is effectively:

```
np.argsort(currDists)[:num_prototypes]
```

...unless I'm missing something? Won't this always find the next smallest distance sum, which in effect would be the above `argsort`? I don't see how the rank order of the values would change given the while block.

phylogeny/prototypeSelection.py
```diff
@@ -448,3 +449,83 @@ def prototype_selection_constructive_protoclass(dm: DistanceMatrix,
...
+    num_prototypes: int
+        Number of prototypes to select for distance matrix.
+        Must be >= 2, since a single prototype is useless.
+        Must be smaller than the number of elements in the distance matrix,
+        otherwise no reduction is necessary.
+
+    Returns
+    -------
+    list of str
+        A sequence holding selected prototypes, i.e. a sub-set of the
```
wasade, line 476: From the return, it looks like this is a subset of the IDs of the matrix not the elements?

phylogeny/prototypeSelection.py
sjanssen2: fixed

phylogeny/prototypeSelection.py
sjanssen2: you are right. I fixed the documentation (of all heuristics)

phylogeny/prototypeSelection.py
sjanssen2: I tend to say that you miss something here. It is more than sorting a list and taking the first X elements. In each iteration, some list elements change and thus the ordering is effected: `currDists -= dm.data[minElmIdx]`
I'd be happy to revise the whole loop with a single line of code, but I don't see how.

wasade: Oh, I see it now. Sorry for the confusion.

sjanssen2:
OK to merge?

PR #53, opened 2017-01-26
[WIP] improvement in Bash scripts

qiyunzhu:
In this PR I demonstrated a new coding style for the Bash scripts (e.g., `run_program.sh`). During last code review, @josenavas and @wasade made multiple valuable comments, based on which I made these modifications. I only updated `run_trex.sh`. If the changes sounds good @ekopylova I will update other scripts as well.

One major problem of the current Bash scripts is that there are too many parameters (e.g., `launch_software.sh` has 29 parameters), and they are not sorted in a constant order. This will impose practical troubles to the users. To solve this problem I replaced the positional parameters with option-argument pairs (e.g., `--input-file /path/to/file.txt`). To achieve this I used [GNU getopt](https://www.gnu.org/software/libc/manual/html_node/Getopt.html), which is installed in most Linux systems by default. I designed a concise and elegant way to do the job, which I am quite proud of.

The second change is adding `set -eu` in the beginning of each script to handle runtime errors.

People also suggested to use GNU time to record time and memory consumption for benchmarking. But @wasade suggested that this applet is buggy in some circumstances, so I didn't use it for the moment.

### DIFF
--- a/benchmark/run_trex.sh
+++ b/benchmark/run_trex.sh
@@ -16,9 +16,9 @@ set -eu
 
 # define parameters using an array
 args=(
-  gene_tree_dir  # hey
-  output_fp      # look
-  verbose        # I can comment each parameter
+  gene_tree_dir        # directory containing gene trees
+  output_fp
+  verbose
   stdout
   stderr
   scripts_dir
--- a/benchmark/run_trex.sh
+++ b/benchmark/run_trex.sh
@@ -9,16 +9,39 @@
 # ----------------------------------------------------------------------------
 
 # usage: run T-REX software
-gene_tree_dir=$1
-output_fp=$2
-verbose=$3
-stdout=$4
-stderr=$5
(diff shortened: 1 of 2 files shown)


benchmark/run_trex.sh
```diff
@@ -9,16 +9,39 @@
...
+
+# -e: script will exit if any command fails
+# -u: force initialization of all variables
+set -eu
+
+# define parameters using an array
+args=(
+  gene_tree_dir  # hey
+  output_fp      # look
+  verbose        # I can comment each parameter
```
ekopylova: Great solution @qiyunzhu !

Could you update the comments and I'll merge. Thanks!

ekopylova (changes requested):

qiyunzhu:
@ekopylova done!

ekopylova (approved):

PR #55, opened 2017-02-01
modularize Bash scripts

qiyunzhu:
In this PR I mainly did three things: 1) moved the command-line argument parsing function from each individual script into `utils.sh`, and have the scripts `source` it. 2) Set undefined arguments to `None`. 3) moved the DIAMOND-calling function from `launch_benchmark.sh` to a separate script `run_diamond.sh`. In the future we can have other scripts call this script as well, and add BLAST or other alternatives in parallel to DIAMOND.


benchmark/run_distance_method.sh
```diff
@@ -10,25 +10,17 @@
...
+source $(dirname "$0")/utils.sh
 args=(
-  working_dir
-  query_species_coding_seqs_fp
-  target_proteomes_dir
-  threads
-  distance_method_install_dir
-  stdout
-  stderr
+    working_dir
```
ekopylova: I think we should remove the distance method from the benchmark (at least for now). It implements an implicit phylogenetic approach, however explicit tools like Ranger-dtl are already very fast & sensitive.

ekopylova (commented):
@qiyunzhu one comment, otherwise looks good

qiyunzhu:
@ekopylova Thanks! I removed distance-methods.

PR #56, opened 2017-02-02
added script for running phylophlan

qiyunzhu:
@ekopylova I added PhyloPhlAn support.

### DIFF
--- a/benchmark/run_phylophlan.sh
+++ b/benchmark/run_phylophlan.sh
@@ -65,6 +65,7 @@ echo "Total user time PhyloPhlAn: ${user_time}" >> $stderr
 echo "Total wall time PhyloPhlAn: ${wall_time}" >> $stderr
 
 # tear down
+rm -rf temp
 cd $PWD
 
 # output genome tree will be:
--- a/benchmark/run_phylophlan.sh
+++ b/benchmark/run_phylophlan.sh
@@ -50,7 +50,7 @@ ln -s ${phylophlan_install_dir}/data
 ln -s ${input_faa_dir} input/genomes
 
 # command
-cmd="${phylophlan_install_dir}/phylophlan.py -i input --nproc $threads --c_dat temp"
+cmd="${phylophlan_install_dir}/phylophlan.py -u input --nproc $threads --c_dat temp"
 if [ "$verbose" == "true" ]
 then
     echo "Command:"$'\n'"  $cmd"
--- a/benchmark/run_phylophlan.sh
+++ b/benchmark/run_phylophlan.sh
@@ -1,0 +1,71 @@
+#!/bin/bash
+
+# ----------------------------------------------------------------------------
+# Copyright (c) 2015--, The WGS-HGT Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+# usage: run PhyloPhlAn to build phylogenetic tree of whole genomes
+
+# prerequisites:
+# one needs to install the development version of PhyloPhlAn:
+#   hg clone https://bitbucket.org/nsegata/phylophlan
+#   cd phylophlan
+#   hg up dev
+# one needs the follow programs installed and callable from $PATH:
+#   usearch
+#   muscle
+#   FastTree
+# one also needs biopython installed
+
+set -eu
+source $(dirname "$0")/utils.sh
+args=(
+    input_faa_dir
+    scripts_dir
+    phylophlan_install_dir
+    threads
+    stdout
+    stderr
+    verbose
+)
+get_args "$@"
+
+if [ "$verbose" == "true" ]
+then
+    echo "Running PhyloPhlAn .."
+fi
+
+# set up
+mkdir -p ${working_dir}/phylophlan
+PWD=$(pwd)
+cd ${working_dir}/phylophlan
+mkdir -p input
+mkdir -p output
+mkdir -p temp
+ln -s ${phylophlan_install_dir}/data
+ln -s ${input_faa_dir} input/genomes
+
+# command
+cmd="${phylophlan_install_dir}/phylophlan.py -i input --nproc $threads --c_dat temp"
+if [ "$verbose" == "true" ]
+then
+    echo "Command:"$'\n'"  $cmd"
+fi
+
+# run PhyloPhlAn and record time
+TIMEFORMAT='%U %R'
+TIME="$( time ($cmd 1>>$stdout 2>>$stderr) 2>&1)"
+user_time=$(echo $TIME | awk '{print $1}')
+wall_time=$(echo $TIME | awk '{print $2}')
+echo "Total user time PhyloPhlAn: ${user_time}" >> $stderr
+echo "Total wall time PhyloPhlAn: ${wall_time}" >> $stderr
+
+# tear down
+cd $PWD
+
+# output genome tree will be:
+# ${working_dir}/phylophlan/output/genomes.tree.nwk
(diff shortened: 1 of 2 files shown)


qiyunzhu:
Also modified the T-REX script.

qiyunzhu:
@sjanssen2 @mortonjt @tkosciol Thanks for your kind offer! Here is the latest PR. It is a Bash script that calls a third-party program (PhyloPhlAn).

benchmark/run_phylophlan.sh
```diff
@@ -0,0 +1,71 @@
...
+fi
+
+# set up
+mkdir -p ${working_dir}/phylophlan
+PWD=$(pwd)
+cd ${working_dir}/phylophlan
+mkdir -p input
+mkdir -p output
+mkdir -p temp
+ln -s ${phylophlan_install_dir}/data
```
tkosciolek, line 49: this line does not have a target

benchmark/run_phylophlan.sh
```diff
@@ -0,0 +1,71 @@
...
+TIMEFORMAT='%U %R'
+TIME="$( time ($cmd 1>>$stdout 2>>$stderr) 2>&1)"
+user_time=$(echo $TIME | awk '{print $1}')
+wall_time=$(echo $TIME | awk '{print $2}')
+echo "Total user time PhyloPhlAn: ${user_time}" >> $stderr
+echo "Total wall time PhyloPhlAn: ${wall_time}" >> $stderr
+
+# tear down
+cd $PWD
+
```
tkosciolek: there is no tearing down part. where do you delete temp, etc.?

benchmark/run_phylophlan.sh
```diff
@@ -0,0 +1,71 @@
...
+# output genome tree will be:
+# ${working_dir}/phylophlan/output/genomes.tree.nwk
```
tkosciolek: maybe you can print it as string in verbose mode?

benchmark/run_trex.sh
```diff
@@ -24,14 +24,19 @@ args=(
 )
 get_args "$@"
 
+$verbose && echo "Running T-REX .."
+cmd="${trex_install_dir}/hgt3.4 -inputfile=${input_file_nwk}"
+$verbose && echo "Command:"$'\n'"  $cmd"
+
+printf "#TREX\n" >> $output_fp
```
tkosciolek, line 31: I think you should use `echo -e`

benchmark/run_trex.sh
```diff
@@ -24,14 +24,19 @@ args=(
...
+touch ${stdout%.*}.total_results.txt
+
 TIMEFORMAT='%U %R'
 total_user_time_trex="0.0"
 total_wall_time_trex="0.0"
```
tkosciolek, line 36: why are you creating those 2 variables if you never use them?

benchmark/run_trex.sh
```diff
@@ -24,14 +24,19 @@ args=(
...
-i=0
 
-printf "#TREX\n" >> $output_fp
-touch ${stdout%.*}.total_results.txt
 # search for HGTs in each gene tree
+i=0
```
tkosciolek, line 39: you're not incrementing this interator anywhere

benchmark/run_trex.sh
```diff
@@ -24,14 +24,19 @@ args=(
...
 for gene_tree in $gene_tree_dir/*.nwk
 do
     gene_tree_file=$(basename $gene_tree)
```
tkosciolek: you're not using `gene_tree_file` anywhere

qiyunzhu:
@tkosciol Thank you! These are valuable comments!

benchmark/run_trex.sh
qiyunzhu: Removed.

benchmark/run_trex.sh
qiyunzhu: It is in line #59.

benchmark/run_trex.sh
qiyunzhu: They are used in lines #56-57.

benchmark/run_trex.sh
tkosciolek: shoot, sorry! GitHub hid those lines from me and I didn't notice. :/

benchmark/run_phylophlan.sh
qiyunzhu: This script will be connected to another script, which hasn't been implemented yet. So I left it as is.

benchmark/run_phylophlan.sh
qiyunzhu: No they are not deleted... We haven't decided how to deal with them yet. @ekopylova What do you think?

benchmark/run_phylophlan.sh
qiyunzhu: `ln` will retain the original file name if there is no target.

benchmark/run_trex.sh
qiyunzhu: This is a good point. Not making any changes but `printf` sends a smell  of C/Perl complication. I replaced all `printf`s with `echo` or `echo -e`. @ekopylova If you have other considerations please let us know.

qiyunzhu:
Also we discussed about two alternative ways of using `$verbose`. Solution A (4 lines):
```
if [ "$verbose" == "true" ]
then
    echo "Command:"$'\n'"  $cmd"
fi
```
Solution 2 (1 line):
```
$verbose && echo "Command:"$'\n'"  $cmd"
```
The latter will require that I modify the `get_args` function in `utils.sh` to force $verbose to be true or false.
Which one do you like? @ekopylova

benchmark/run_trex.sh
qiyunzhu: @tkosciol I was about to change, but I noticed some issues. See this [post](http://stackoverflow.com/questions/525872/echo-tab-characters-in-bash-script) for people who prefer `printf` over `echo`.
For example, originally it was:
```
printf "$i\t$gene_number\t" >> $output_fp
```
If I change it into `echo`, it becomes:
```
echo $i$'\t'$gene_number$'\t' >> $output_fp
```
Sounds bit awkward. What do you think?

benchmark/run_trex.sh
qiyunzhu: No problem :)

benchmark/run_phylophlan.sh
mortonjt: :+1: for deleting tmp.  You want to avoid bloat, since that filesystems will crash if there are too many files.

qiyunzhu: @mortonjt Thanks! Deleted. By the way, how do you think of the $verbose issue (see end of the page)?

qiyunzhu:
Would anyone like to give some final comments and/or merge? @mortonjt @tkosciol @ekopylova

benchmark/run_phylophlan.sh
```diff
@@ -0,0 +1,72 @@
...
+# ----------------------------------------------------------------------------
+
+# usage: run PhyloPhlAn to build phylogenetic tree of whole genomes
+
+# prerequisites:
+# one needs to install the development version of PhyloPhlAn:
+#   hg clone https://bitbucket.org/nsegata/phylophlan
+#   cd phylophlan
+#   hg up dev
+# one needs the follow programs installed and callable from $PATH:
```
ekopylova, line 18: any specific versions?

ekopylova (commented):
@qiyunzhu one comment, otherwise looks good!

benchmark/run_phylophlan.sh
qiyunzhu: This is the development version Nicola and Francesco are working on. After talking to them they showed me these commands. They said depite not having a version number, this version is pretty stable. They expect to resume the development from April to add some major improvements. @ekopylova

ekopylova:
@qiyunzhu  thanks!

qiyunzhu:
@ekopylova Thank you for merging!

PR #59, opened 2017-02-11
DarkHorse2 automated installation script

qiyunzhu:
Completed. Feel free to review and merge.

Among all applications, [DarkHorse](https://github.com/spodell/Darkhorse2)'s installation is notably difficult. It relies on the **MySQL** database service running in the system. Previous, a system-wide installation of DarkHorse 1.5 is in `/opt/darkhorse` of Barnacle, and it is bond to a copy of NCBI nr as of early 2015. Often, we have to ask Jeff to launch the MySQL service before we can run DarkHorse. Any modifications will probably be thru Jeff, too. This is very inconvenient. I think that the average end users may or may not have the expertise / privilege to install MySQL.

To solve this problem, I figured out a way to install MySQL to a user directory - **without admin privilege** - and run it like a sandbox. Here is the code I wrote. The MySQL part was tested and working. But the DarkHorse part was but envisioned, because I noticed that the way DarkHorse connects MySQL may need some modifications.

Specifically, the sandbox MySQL is connected by shell command:
`mysql --socket /path/to/mysql.sock -u root --skip-password`

The `--skip-password` parameter is not important, but the socket is critical, since it is in a user directory.

Sheila @spodell , is it be possible that DarkHorse can call MySQL like this?

Also pings @ekopylova @wasade @sjanssen2 in case you are interested.

===========

I think I solved the socket issue.


qiyunzhu:
The script is completed. Tested and working. I expect that it will work in any Linux or Mac computers, without admin privilege.

benchmark/run_phylomizer.sh
```diff
@@ -99,7 +99,7 @@ do
 
     # report the best aa substitution model and the -log likelihood of the tree
     best_tree=$(head -n1 input.tree.phyml.rank.ml)
-    best_model=$(echo ${best_tree} | cut -f1)
+    best_model=$(echo "${best_tree}" | cut -f1)
```
tkosciolek, line 102: Just to be safe, wouldn't you want to specify a delimiter, i.e. `cut -d ' ' -f1`

benchmark/utils.sh
```diff
@@ -33,7 +33,7 @@ function get_args() {
     for arg in ${args[@]}
     do
         # set undefined arguments to None
-        if [ -z "$(echo ${!arg})" ]
+        if [ -z "${!arg+x}" ]
```
tkosciolek, line 36: what is `x` doing here, where does it come from?

darkhorse/install_darkhorse.sh
```diff
@@ -0,0 +1,169 @@
...
+# usage: automate the installation of DarkHorse2
+# this script will install:
+#   MySQL 5.7.18
+#   NCBI nr, taxdump and prot.accession2taxid (up-to-date)
+#   DarkHorse 2.0_rev06
+# the whole system (including MySQL) will run in a local directory and won't
+# affect anything outside
+
+set -eu
+appdir=$(readlink -m $1)
```
tkosciolek: add usage info where/how we define `$1`

darkhorse/install_darkhorse.sh
```diff
@@ -0,0 +1,169 @@
...
+
+# download and compile MySQL source code
+mkdir -p $appdir/mysql
+wget https://dev.mysql.com/get/Downloads/MySQL-5.7/mysql-boost-5.7.18.tar.gz
```
tkosciolek: wouldn't it be safer to use `rsync`or `curl`? doen't really matter, though

darkhorse/install_darkhorse.sh
```diff
@@ -0,0 +1,169 @@
...
+tar xvf mysql-boost-5.7.18.tar.gz
```
tkosciolek: `tar -zxvf` should make it version-proof. Are you sure you can unpack `tar.gz` without `-z` flag?

darkhorse/install_darkhorse.sh
```diff
@@ -0,0 +1,169 @@
...
+sql_mode       = NO_ENGINE_SUBSTITUTION
+[mysqld_safe]
+log-error      = $appdir/mysql/data/mysql.err
+pid-file       = $appdir/mysql/mysql.pid
+" > $appdir/mysql/etc/my.cnf
+
+export PATH=$appdir/mysql/bin:$PATH
+
+# initialize MySQL serivce
+mysqld --initialize-insecure
```
tkosciolek, line 77: I assume it immediatly goes into background, that's why you need `sleep 60`?

darkhorse/install_darkhorse.sh
```diff
@@ -0,0 +1,169 @@
...
+
+# download NCBI database
+mkdir -p $appdir/ncbi
+cd $appdir/ncbi
+wget ftp://ftp.ncbi.nih.gov/blast/db/FASTA/nr.gz
+wget ftp://ftp.ncbi.nih.gov/pub/taxonomy/taxdump.tar.gz
+wget ftp://ftp.ncbi.nih.gov/pub/taxonomy/accession2taxid/prot.accession2taxid.gz
+gunzip nr.gz
+mkdir -p taxdump
+tar xvf taxdump.tar.gz -C taxdump
```
tkosciolek: same as above... `zxvf`

darkhorse/install_darkhorse.sh
```diff
@@ -0,0 +1,169 @@
...
+export PERL_MM_USE_DEFAULT=1
+perl -MCPAN -e 'install App::cpanminus'
+cpanm DBI DBD::mysql
+
+# download DarkHorse 2-rev06
+dhv=rev06
+mkdir -p $appdir/darkhorse
+cd $appdir/darkhorse
+wget https://github.com/spodell/Darkhorse2/archive/$dhv.tar.gz
+tar xvf $dhv.tar.gz
```
tkosciolek: same here... `zxvf`

tkosciolek (commented):
Looks good, except where highlighted.
Please note, that I did not check if it works on my system, just checked code clarity and what steps are executed.

qiyunzhu:
Hi @tkosciol thanks for such careful commenting! Let me test some of the issues you proposed and get back to you soon!

darkhorse/install_darkhorse.sh
qiyunzhu: You are right! I will change into:
```
url=https://dev.mysql.com/get/Downloads/MySQL-5.7/mysql-boost-5.7.18.tar.gz; wget $url || curl -O $url
```

darkhorse/install_darkhorse.sh
qiyunzhu: Makes sense! Added.

darkhorse/install_darkhorse.sh
qiyunzhu: Thanks! I didn't know that before. `tar` can unpack `tar.gz` without `-z`.

darkhorse/install_darkhorse.sh
qiyunzhu: You are correct. Initiating MySQL takes some time in the background.

darkhorse/install_darkhorse.sh
qiyunzhu: Changed.

darkhorse/install_darkhorse.sh
qiyunzhu: Changed.

benchmark/utils.sh
qiyunzhu: I learned this trick from here: https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash

benchmark/run_phylomizer.sh
qiyunzhu: The delimiter is `\t`, which is `cut`'s default delimiter. If one specifies it, the command will be as awkward as `cut -d$'\t' -f1`

qiyunzhu:
Hello @tkosciol I addressed your comments. Would you like to take another look?

darkhorse/install_darkhorse.sh
```diff
@@ -0,0 +1,178 @@
...
+
+# automated installation script for DarkHorse2
+# this script will install:
+#   MySQL 5.7.18
+#   NCBI nr, taxdump and prot.accession2taxid (up-to-date)
+#   DarkHorse 2.0_rev07
+# the whole system (including MySQL) will run in a local directory and won't
+# affect anything outside
+# usage:
+#   bash install_darkhorse.sh /target/directory/to/install
```
mortonjt, line 19: are these comments necessary?

mortonjt (commented):
This is quite insane to put together.  Given that this is already working, I think this is good to go!

darkhorse/install_darkhorse.sh
qiyunzhu: @mortonjt Thanks! I think more documentation doesn't hurt!

PR #60, opened 2017-02-18
added taxon sampling script

qiyunzhu:
Here I added a new script to sample taxa from DIAMOND hit table. This script is very important and tricky. Actually, base on my previous experience in HGT discovery, taxon sampling is the most critical part of the workflow, and hasn't been sufficiently studies. Therefore this part deserves extensive further research in our project. However, for now, I only implemented a very simple script that undifferentially sample all taxa found in the hit table. For more details please see the docstring of the code.

I also updated the Bash script to match the new versions of DIAMOND.


qiyunzhu:
Also tested with Python 3.6. It worked!

benchmark/launch_benchmark.sh
```diff
@@ -93,23 +97,34 @@ then
...
+python ${scripts_dir}/sample_taxa.py --hit-table-fp CrDC.m8 ${hit_table_fp} \
+                                     --prot2tax-dict-fp ${prot2tax_dict_fp} \
+                                     --output-taxa-fp ${working_dir}/sampled_taxa.txt
+# soft link selected protein sequence files to output directory
+mkdir -p ${working_dir}/sampled_faa
+ln -s ${species_faa_fp} ${working_dir}/sampled_faa/query.faa
+while read taxon
+do
+    ln -s ${database_faa_dir}/$taxon.faa ${working_dir}/sampled_faa/$taxon.faa
+done < ${working_dir}/sampled_taxa.txt
```
tkosciolek, line 126: I guess it works, but that's a really nasty way of passing data into the loop.
wouldn't it be cleaner if you did a for loop, e.g.:
`for taxon in $(cat ${working_dir}/sampled_taxa.txt)`
or
`cat ${working_dir}/sampled_taxa.txt | while read taxon` ?

benchmark/sample_taxa.py
```diff
@@ -0,0 +1,122 @@
...
+          output_taxa_fp):
+    """ Select taxa for subsequent analyses
+    """
+    sample_taxa(hit_table_fp,
+                prot2tax_dict_fp,
+                output_taxa_fp)
+
+
+if __name__ == "__main__":
+    _main()
```
tkosciolek: This script looks good.
I'd maybe change the function name from `_main()` to something more descriptive.
Also, is there a specific reason why you're writing `taxa` to file in `sample_taxa()`, instead of returning it and then writing it to file in `main()`.
In my example, if you just call the function, you get a data structure and if you call the script, you get a file.

tkosciolek (commented):
Overall looks good! Just 2 minor comments, so it's almost ready to merge.

benchmark/launch_benchmark.sh
qiyunzhu: Hi Tomasz, thank you! I was about to accept your suggestion, but I did a bit more Googling and figured that your way may not work as expected. See this [article](http://www.compciv.org/topics/bash/loops/). Besides, all articles I read about looping file lines used `while read ... done < file`.

benchmark/sample_taxa.py
qiyunzhu: This is a good idea! It will cause some **coverage decrease** though. If you don't mind, I will change it to return + write.

qiyunzhu: Btw, "main" was the convention of the WGS-HGT codebase, so I didn't change it. Otherwise I would love a more meaningful name!

benchmark/launch_benchmark.sh
tkosciolek: Interesting, I didn't know that!
Thanks for looking this article up!

benchmark/sample_taxa.py
tkosciolek: I'm just suggesting. I'd say the code is good to go, as long as you're happy with it. I don't see any major problems.:)

qiyunzhu: Cool! Let me change into your way (return + write). I like it better!

benchmark/launch_benchmark.sh
qiyunzhu: You are welcome!

PR #61, opened 2017-02-18
refreshed orthofinder support to latest generic version

qiyunzhu:
**OrthoFinder** is used for orthology identification, which is essential for subsequent gene family tree building. However, this program only supports Python 2. Previously, a special version of OrthoFinder 0.4 which supports Python 3 was obtained from the developers. It is outdated however. In order to run up-to-date versions of OrthoFinder (now 1.1.4), a separate Python 2 conda environment is created, and is called from a new Bash script (like other external programs), instead of directly called from the Python scripts.

Note: OrthoFinder requires BLAST+ for homology search. Meanwhile, it has an option that allows the user to provide pre-computed BLAST results. Therefore, there is room for hacking this process by providing DIAMOND results. It is to be further explored.

Update: I did this hack. So now OrthoFinder is married to DIAMOND.

Note: This PR is NOT dependent on the last PR ([#60](https://github.com/biocore/WGS-HGT/pull/60)).

PS: There really **isn't any unit test added or removed**. The coverage decrease is simply because some redundant codes are removed.

@ekopylova @wasade


qiyunzhu:
PS: There really isn't any unit test added or removed. The coverage decrease is simply because some redundant codes are removed.

mortonjt:
Are you planning on making this tool uninstallable?  If so, you'll need to make this either strictly Python 2 or Python 3.  Since we are in the process of making all of our tools Python 3 compatible, it'll make it difficult to directly interface with our tools in the lab.  So it is not recommended to go the Python 2 route.

If you decide to make everything Python 3 compatible, there are a few possible options

1. Copy over OrthoFinder to the WGS-HGT and make it Python 3.  Looks like there is very little code in the main repository, so it should be very fast to do that.  However, you'd have to kill the BSD license in favor of GPL, since you would be copying GPL code.
2. Push in a PR into OrthoFinder to make it Python 3 compatible.  Definitely the most preferred route, and it looks like it is being actively maintained.

qiyunzhu:
@mortonjt Thanks for your comments! To everyone else, Jamie and I discussed offline about the dependency issue. The reality is that WGS-HGT has whole loads of dependency issues. Python 2 is not the worst (see [here](https://github.com/biocore/WGS-HGT/pull/59) for another example). Now that I managed to figure our an automated solution to run both Python 2 and 3 programs in the same pipeline, I think we can live with it. (Actually I figured out two solutions. If you want to know please leave message). There is a long way to go before making WGS-HGT user friendly.

This is how the dependencies looked like 6 months ago:

screen shot 2017-02-18 at 6 09 46 pm

This is how they look like now:

```
conda create -n wgshgt python=3.5
source activate wgshgt
conda install click biopython
conda install -c biocore scikit-bio
conda install -c bioconda blast diamond fasttree mafft mcl muscle prodigal=2.6.2 raxml t_coffee trimal
source deactivate
conda create -n wgshgt_py2 python=2.7 -c bioconda orthofinder
# plus some other programs...
```

mortonjt:
Note that the IntervalMetadata is now merged into scikit-bio.  So we don't need to rely on my conda channel anymore.

qiyunzhu:
@mortonjt Yes I noticed and already removed it.

qiyunzhu:
@mortonjt @ekopylova @wasade Anyone would like to take a look? Thanks!

RNAer:
👍

mortonjt (approved):

qiyunzhu:
Thank you bros! @mortonjt @RNAer @tkosciol

.travis.yml
```diff
@@ -13,60 +13,28 @@ before_install:
...
-  # Install PhyloNet
+  ## Install PhyloNet
   # - wget http://bioinfo.cs.rice.edu/sites/g/files/bxs266/f/kcfinder/files/PhyloNet_3.5.7.jar -O java-files/PhyloNet_3.5.7.jar --no-check-certificate
 install:
   # Install Python packages and some external programs using conda + pip
   - conda create -n test_env python=$PYTHON_VERSION --file ci/conda_requirements.txt
   - source activate test_env
+  - conda install -c bioconda diamond
   - pip install -r ci/pip_requirements.txt
   - pip install .
```
serenejiang, line 35: what are you installing with this line?

serenejiang (commented):
Great!

.travis.yml
qiyunzhu: Hi @serenejiang Thanks for taking a look at this code!
This particular command installs "diamond" from conda channel "bioconda". [Diamond](https://github.com/bbuchfink/diamond) is a program for sequence similarity search.

PR #62, opened 2017-02-24
added sample genes script

qiyunzhu:
I added a new Python script "**sample_genes.py**". This is to select proper gene families based on OrthoFinder results for subsequent phylogenetic reconstruction and phylogeny-based HGT detection. There are three functions in total. To simplify code review I took away two functions and only left the first function. I also did a few other tune-ups.

I also upgraded the way DIAMOND is called, by adding some search cutoff parameters. This changed a few files, but actually this is a very simple change.

The only major block of code to review is merely **sample_genes.py**.


qiyunzhu:
Hello @sjanssen2 the Travis check was finally run. All look good. Could you please review? Thanks!

sjanssen2:
Hi Qiyun,

sure I will. Can you use the time I need to review to do me the favour 
and review this here: https://github.com/biocore/conda-recipes/pull/90

Thanks,
Stefan

qiyunzhu:
Hello @sjanssen2 , no problem! I just reviewed you PR [#90](https://github.com/biocore/conda-recipes/pull/90). It is my pleasure to help checking your code. Please don't hesitate to let me know if you need more reviews in the future.

benchmark/tests/test_sample_genes.py
```diff
@@ -0,0 +1,63 @@
...
+        self.assertRaises(ValueError, sample_genes, self.ortho_groups_fp,
+                          min_taxa_cutoff=-1)
+
+    def test__main(self):
+        params = ['--ortho-groups-fp', self.ortho_groups_fp,
+                  '--min-taxa-cutoff', 4.0]
+        res = CliRunner().invoke(_main, params)
+        self.assertEqual(res.exit_code, 0)
+        self.assertEqual(res.output, 'Number of gene families sampled: 6.\n')
+
```
sjanssen2, line 60: If you look at the coverall report for  /benchmark/simulate_hgts.py  you will see that some lines are uncovered by your tests, especially raising exceptions. Please add tests that check if those exceptions are raised, e.g. like 
```
with self.assertRaisesRegex(IOError, 'Cannot write to file'):
       yourFunction('/dev/', p)
```

benchmark/launch_benchmark.sh
```diff
@@ -114,17 +128,37 @@ fi
 
 ## step 2:
 ##    select reference genomes containing homology with query genome
-python ${scripts_dir}/sample_taxa.py --hit-table-fp CrDC.m8 ${hit_table_fp} \
+python ${scripts_dir}/sample_taxa.py --hit-table-fp ${hit_table_fp} \
```
sjanssen2, line 131: what happend to CrDC.m8 ?

benchmark/tests/test_sample_genes.py
```diff
@@ -0,0 +1,63 @@
...
+    """ Test for sample_taxa.py """
+
+    def setUp(self):
+        """ Set up working directory and test files
+        """
+        # test output can be written to this directory
+        self.working_dir = mkdtemp()
+
+        # test data directory
+        datadir = join(dirname(realpath(__file__)), 'data', 'simplified')
```
sjanssen2, line 29: don't we have this in skbio as from skbio.util import get_data_path ?

benchmark/sample_genes.py
```diff
@@ -0,0 +1,90 @@
...
+        one or more homologs of the query gene. default: 10.0
+
+    Return
+    ------
+    dict of dict of set of str
+        { ogid : { taxon : set(protein(s)) }
+            ogid: ID of the orthologous group (assigned by OrthoFinder)
+            taxon: ID of the taxon (genome)
+            protein: ID of the protein
+
```
sjanssen2, line 39: you should add a section 'Raises' to document which exceptions can be raised by your function.

benchmark/sample_genes.py
```diff
@@ -0,0 +1,90 @@
...
+    taxa = ogs.columns
+    # set minimum number of taxa cutoff
+    if min_taxa_cutoff > 1:
+        min_taxa_cutoff = int(min_taxa_cutoff)
+    elif 0 < min_taxa_cutoff <= 1:
+        min_taxa_cutoff = int(len(ogs.columns) * min_taxa_cutoff)
+    else:
+        raise ValueError('Error: %s is not a valid minimum taxa cutoff.'
+                         % min_taxa_cutoff)
+    genes = {}
```
sjanssen2, line 65: I am not sure if it's good for readability, but why not using more power of pandas:

```
x = ogs.applymap(lambda x: set(x.split(', ')) if pd.notnull(x) else {})
x[(x.applymap(len).sum(axis=1) >= min_taxa_cutoff) & (x['query'] != {})]
genes = x.T.to_dict()
```

sjanssen2 (commented):
only minor comments. I haven't executed the code, thus you need to make sure that everything works as expected.

benchmark/launch_benchmark.sh
qiyunzhu: That was a typo. So I removed it.

benchmark/sample_genes.py
qiyunzhu: Good point! Sorry I forgot. Added.

benchmark/tests/test_sample_genes.py
qiyunzhu: Yes we do. But I love this native Python solution. Why scikit-bio has to include some file system functions?

benchmark/tests/test_sample_genes.py
qiyunzhu: Well, this is a historical issue. I wasn't the one that wrote most of `simulate_hgts.py` and a few other scripts. I found that these codes are pretty deliberate and sophisticated, and it isn't time-friendly to fully understand the logic, not mentioning adding tests. This could be fixed in the future.

benchmark/launch_benchmark.sh
```diff
@@ -68,6 +68,16 @@ args=(
...
 
+    ## sequence similarity search parameters (BLAST / DIAMOND)
+    # maximum E-value cutoff
+    evalue
+    # minimum identity cutoff (%)
+    pident
+    # minimum query coverage cutoff (%)
+    qcovs
+    # maximum number of target sequences to report
+    max_target_seqs
```
mortonjt, line 79: What are you trying to do here?

benchmark/tests/test_sample_genes.py
mortonjt: This is just for testing purposes.  Personally, I use it often for my own solutions since it is pretty easy / readable.
Since skbio is already a dependency, using it here is not an issue.  But I personally don't have preferences wrt this.

mortonjt:
Is there any interest in pushing this bash code to snakemake in the future?  This workflow could be seriously simplified in the process.  In addition, you get out of the box parallelization while making it easy to pop on a wide variety of compute environments, including slurm and torque.

benchmark/sample_genes.py
qiyunzhu: @sjanssen2 and I discussed in person. We decided to keep the original way.

benchmark/launch_benchmark.sh
qiyunzhu: These are but standard BLAST (DIAMOND) parameters. I added those so that the user can customize their BLAST search.

qiyunzhu:
@mortonjt Hi Jamie, thanks for your comments! Yes I do have the desire to migrate the whole thing into snakemake or other advanced pipelining platforms in the future. I think that one huge advantage of snakemake is that we can code Bash and Python simultaneously, which can simplify the current WGS-HGT (Bash scripts + Python scripts) quite a bit.

PR #63, opened 2017-02-24
added write sampled genes function

qiyunzhu:
Here I added a new function to **sample_genes.py**. The function of the function is simple: write sampled gene list and sequences to external files. In this PR a few test data files are added, but the code that actually needs to be reviewed is merely sample_genes.py and test_sample_genes.py.


mortonjt:
Just as a general comment it may be worthwhile to take a look at python [generators](https://wiki.python.org/moin/Generators
).  It could drastically simplify your code by eliminating complex data-structures.  In addition, it would allow for large memory objects to be shuffled around, since lines in a file are retrieved only when they need to be.  Less memory use = faster code.

Of course this is optional for this PR, but it is something to keep in mind for future refactors.

benchmark/sample_genes.py
```diff
@@ -74,6 +76,72 @@ def sample_genes(ortho_groups_fp,
...
+                    prots[taxon][prot].add(ogid)
+                else:
+                    prots[taxon][prot] = set([ogid])
+    # match taxa with faa filenames
+    #   so complicated because OrthoFinder trims off the version number from
+    #   an NCBI-style accession (e.g., GCF_012345.1 becomes GCF_012345)
+    taxon2file = {}
+    for file in os.listdir(input_faa_dir):
+        if file.endswith('.faa'):
+            taxon = file.split('.')[0]
```
mortonjt: I'd recommend taking a look at [splitext](http://stackoverflow.com/a/541394/1167475)

benchmark/sample_genes.py
```diff
@@ -74,6 +76,72 @@ def sample_genes(ortho_groups_fp,
...
+    for ogid in sorted(genes):
+        members = []
+        with open(os.path.join(output_fa_dir, '%s.fa' % ogid), 'w') as f:
+            for taxon in sorted(genes[ogid]):
+                # restore taxon name from OrthoFinder-crippled form
+                trutax = taxon2file[taxon][:-4]
+                for prot in sorted(genes[ogid][taxon]):
+                    if prot in seqs:
+                        # sequence IDs are like: taxon|protein
+                        f.write('>%s|%s\n%s\n' % (trutax, prot, seqs[prot]))
```
mortonjt, line 139: I would strongly recommend taking advantage of the metadata internals in [scikit-bio](http://scikit-bio.org/docs/0.5.1/generated/skbio.sequence.Sequence.metadata.html#skbio.sequence.Sequence.metadata) if possible.

This will make your code so much easier to maintain, since can take advantage of the existing standards in skbio, rather than inventing your own file-formats.  From personal experience, inventing and maintaining file formats is a nightmare.

benchmark/sample_genes.py
```diff
@@ -74,6 +76,72 @@ def sample_genes(ortho_groups_fp,
...
+def write_genes(genes,
+                input_faa_dir,
+                output_fa_dir,
+                output_genes_fp):
+    """ Write protein sequences of selected gene families to external files
+
+    Parameters
+    ----------
+    genes : dict of dict of set of str
+        { ogid : { taxon : set(protein(s)) }
```
mortonjt, line 88: Is it possible to refactor the code previously merged to take advantage of scikit-bio objects (i.e. sequence object).  Passing around dict of dict of set of str is definitely not ideal ...

benchmark/sample_genes.py
```diff
@@ -74,6 +76,72 @@ def sample_genes(ortho_groups_fp,
...
+            for prot in genes[ogid][taxon]:
+                if prot in prots[taxon]:
+                    prots[taxon][prot].add(ogid)
+                else:
+                    prots[taxon][prot] = set([ogid])
+    # match taxa with faa filenames
+    #   so complicated because OrthoFinder trims off the version number from
+    #   an NCBI-style accession (e.g., GCF_012345.1 becomes GCF_012345)
+    taxon2file = {}
+    for file in os.listdir(input_faa_dir):
```
mortonjt: this is a reserved keyword - would recommend changing this to something else like `fname`

qiyunzhu:
Hi @mortonjt Thank you! Let me read about Python generators!

benchmark/sample_genes.py
mortonjt: Or if you are using scikit-bio objects, make sure to document it in the Parameters section

benchmark/sample_genes.py
```diff
@@ -74,6 +76,72 @@ def sample_genes(ortho_groups_fp,
...
+            if taxon in prots:
+                taxon2file[taxon] = file
+    # read protein sequences
+    seqs = {}
+    for taxon in taxon2file:
+        for seq in io.read(os.path.join(input_faa_dir, taxon2file[taxon]),
+                           format='fasta'):
+            id = seq.metadata['id']
+            if id in prots[taxon]:
+                seqs[id] = str(seq)
```
mortonjt, line 128: Read and write operations should be moved to separate modules if possible.  It'll be easier to test / maintain in the future.

benchmark/sample_genes.py
```diff
@@ -74,6 +76,72 @@ def sample_genes(ortho_groups_fp,
...
+    prots = {}
+    for ogid in genes:
+        for taxon in genes[ogid]:
+            if taxon not in prots:
+                prots[taxon] = {}
+            for prot in genes[ogid][taxon]:
+                if prot in prots[taxon]:
+                    prots[taxon][prot].add(ogid)
+                else:
+                    prots[taxon][prot] = set([ogid])
```
mortonjt, line 111: I'd reconsider the data structure required to carry out this task.  The transformations performed here aren't very clear, and seem a bit complex.

mortonjt (commented):
Take my comments with a grain of salt.  They are optional, but could make code management in the future much easier on yourself.

benchmark/sample_genes.py
qiyunzhu: Thanks! The problem is that OrthoFinder isn't doing a clean way. What they get as a stem file name is actually the first part of the whole file name (e.g., `this.is.my_file.txt` becomes `this`), unfortunately. I hope they do `splitext` instead.

benchmark/sample_genes.py
qiyunzhu: I like scikit-bio's metadata internals! Here the output file format has to be FASTA, because they are to be fed to downstream phylogenetics programs (Phylomizer, and others). The only thing I "invented" here is that the sequence ID is changed from "protein" into "taxon|protein". Is there a standard in scikit-bio already to handle this case?

benchmark/sample_genes.py
qiyunzhu: It is complicated! The logic is that: The FASTA files to be read (see a few line below), which are the reference proteome sequences, are big. It may be inefficient to load them all into memory. Therefore, it is important to sequentially read each line and check if this is the sequence I want, if yes then load it into memory. To do this test, the code needs two identifiers: taxon, and protein (taxon is the FASTA filename and protein is the sequence name). Therefore I translated the original data structure:
`{ gene : { taxon : { set of protein(s) } } }`
into:
`{ taxon : { protein : gene } }`
Does that sound reasonable?

benchmark/sample_genes.py
qiyunzhu: I have some experience with scikit-bio's Sequence object, but I couldn't think of a way to take advantage of it, otherwise I already did (see `io` below). Do you have a clue for it?
Maybe **pandas** instead of **scikit-bio** is the right package for this case. Still I couldn't think of a solution with pandas. Therefore I did the most native way.

benchmark/sample_genes.py
qiyunzhu: There is some sort of the flavor you suggested in WGS-HGT. For example, there are `reformat_input.py` and `parse_output.py` to handle the input / output files for various HGT detection programs. However, I increasingly feel that this setting is awkward. Different program produces very different output files, and we need to parse them for very different purposes. Therefore, it is not possible to let one Python script parse all program output files and generate a uniform type of result.
I think a manageable way is to write one wrapper script (Bash or Python) for each third-party program. This will guarantee modularity of the plugins. Meanwhile we can define several categories of programs (e.g., sequence similarity searchers (BLAST, DIAMOND, VSEARCH), multiple sequence aligners (MAFFT, MUSCLE, T-COFFEE), etc.) and set up rules for the format of input / output files for each category. There will be quite a few categories, though...

benchmark/sample_genes.py
qiyunzhu: Thanks for pointing this out! Changed.

benchmark/sample_genes.py
```diff
@@ -74,6 +76,72 @@ def sample_genes(ortho_groups_fp,
...
+    #   so complicated because it is optimized for subsequent filesystem I/O
+    prots = {}
+    for ogid in genes:
+        for taxon in genes[ogid]:
+            if taxon not in prots:
+                prots[taxon] = {}
+            for prot in genes[ogid][taxon]:
+                if prot in prots[taxon]:
+                    prots[taxon][prot].add(ogid)
+                else:
```
RNAer, line 110: you don't need this `else`:
````
a = set()
a.add('a')
a.add('a')
a  # -> {'a'}
````

RNAer (changes requested):

benchmark/sample_genes.py
qiyunzhu: @RNAer Thanks! I didn't get you though... Can you clarify?

RNAer (approved):

PR #65, opened 2017-03-11
added include function to prototype selection

qiyunzhu:
I added a new parameter "include" to the prototype selection functions implemented by @sjanssen2 . This is to force the function to include a mandatory core set of data points (here referring to taxa that are biologically important) when selecting prototypes.

To scale the algorithms to large amount of data sets, optimization is crucial. However, I found that the introduction of this new parameter substantially slows down the speed. Here is the question:

* Without the new parameter, the program finds the smallest element in a Numpy array using the `argmin` method.

* With the new parameter, the program finds the smallest element in a Numpy array **whose index is not present in another Python set**.

I chose to use the `argsort` method to sort the array, then to iterate from the smallest end until reaching an element whose index in not in the set.

In my benchmark against a distance matrix of 1000 taxa, without the parameter, it took 4.5-5 sec; with the parameter, it took 90 sec. There is a 20-fold decrease in speed. I guess that the parameter made the O(_n_) into O(_n_ log _n_).

Would anyone think of a better implementation? Or this is already the best possible solution? @wasade @mortonjt @RNAer


sjanssen2:
wow, that is a dramatic slowdown. I used to compute 5k out of 30k EMP samples within 6 seconds with this function!

phylogeny/prototypeSelection.py
```diff
@@ -511,7 +519,9 @@ def prototype_selection_constructive_maxdist(dm: DistanceMatrix,
     # the element to remove first is the one that has smallest distance to all
     # other. "Removing" works by tagging its distance-sum as infinity. Plus, we
     # decrease the number of available elements by one.
-    minElmIdx = currDists.argmin()
+    minElmIdx = next(x for x in currDists.argsort()
```
josenavas, line 522: I think the slow down comes from the fact that you're sorting, specially in line 534, that the sorting is inside a for loop (which means that you're sorting multiple times). Instead of argsort use [argmin](https://docs.scipy.org/doc/numpy/reference/generated/numpy.argmin.html) since you're only interested in the minimum one.

The for loop is also performing a filtering (due to the include parameter). Instead of filtering constantly the entire dataset, perform a first pass on the data to filter the data and just work on the filtered data.

qiyunzhu: @josenavas Thank you for reading and thinking of my question! Very valuable comments!

You are right that multiple argsort is what slows down the program. I don't know if argmin can return the index **that is not present in another Python set**. Maybe there is some workaround to achieve this?

@sjanssen2 also thought about the _filter-then-loop_ idea. However, we discussed and figured that this will result in statistically less robust prototypes. The "include" list of elements serve as a core set, from which the prototypes will grow. Leaving them behind will alter the way prototypes grow.

sjanssen2: Wait wait wait. We discussed that and agreed that results will most likely be different, but we have not evaluated if the differences are significant. Therefore, we need to run some evaluation. You should be able to do so now - even if slow - with both versions. Just pick 50 out of 1000 elements with 50 preset (so 100 total) and check the sumdist() values for both results.

qiyunzhu: @sjanssen2 Let me test then :)

qiyunzhu: @sjanssen2 here are the benchmark result: with the test dataset you already implemented in the unit test:

2 cores, 5 prototypes from 20 elememts:
no-core: 5.4494, filter-while-loop: 5.3082, filter-then-loop: 5.2166

3 cores, 10 prototypes from 100 elements:
no-core: 26.8818, filter-while-loop: 26.7458, filter-then-loop: 26.5841

The result suggests that filter-then-loop is a fast heuristic solution at the cost of reduced sum of distances.

sjanssen2: Cool! I'd like to see the same stats for larger numbers, but I guess that is currently too expensive. The remaining question is: is this difference significant? Keep in mind, that the current function is already heuristic - thus we already made the sacrifice of optimality for speed.

sjanssen2: I think I found a more elegant / fast way to implement this functionality. I also thought that the name 'seedset' might be more appropriate, thus I renamed the parameter accordingly. See my PR #66 
The idea is to artificially assign highest distance values to the elements of the seedset and than run the existing algorithm unchanged. Thus, it will have the same runtime for empty seedsets and only very minor additional (linear) runtime for seedsets.

wasade:
Has the code been profiled?

sjanssen2:
How would I do a profiling in Python?
Please turn your focus on PR #66 instead of this one here. It should retain original speed - at least from doing tests via %timeit on 30k EMP matrix

qiyunzhu:
Let me close this PR since Stefan's #66 provided a much faster solution.

PR #66, opened 2017-03-13
support for 'seedsets' in destructive_maxdist heuristic:

sjanssen2:
Qiyun want's to ensure a given set of IDs to be included in the resulting prototype selection set. Thus, we need to violate the global objective function and sneak those IDs in first.

### DIFF
--- a/phylogeny/prototypeSelection.py
+++ b/phylogeny/prototypeSelection.py
@@ -451,7 +451,8 @@ def prototype_selection_constructive_pMedian(dm, num_prototypes):
     return [dm.ids[idx] for idx in prototypes]
 
 
-def prototype_selection_destructive_maxdist(dm, num_prototypes):
+def prototype_selection_destructive_maxdist(dm, num_prototypes,
+                                            seedset=set([])):
     '''Heuristically select k prototypes for given distance matrix.
 
        Prototype selection is NP-hard. This is an implementation of a greedy
@@ -469,6 +470,12 @@ def prototype_selection_destructive_maxdist(dm, num_prototypes):
         Must be >= 2, since a single prototype is useless.
         Must be smaller than the number of elements in the distance matrix,
         otherwise no reduction is necessary.
+    seedset: set(ids)
+        A set of element IDs that are preferably selected for the resulting
+        set. All get selected first, if num_prototype >= len(seedset),
+        otherwise a random sub-selected of seedset is returned.
+        Warning: It will most likely violate the global objective function to
+        pre-select elements.
 
     Returns
     -------
@@ -485,7 +492,7 @@ def prototype_selection_destructive_maxdist(dm, num_prototypes):
 
     Notes
     -----
-    Timing: %timeit -n 100 prototype_selection_constructive_maxdist(dm, 100)
+    Timing: %timeit -n 100 prototype_selection_destructive_maxdist(dm, 100)
             100 loops, best of 3: 2.1 s per loop
             where the dm holds 27,398 elements
     function signature with type annotation for future use with python >= 3.5:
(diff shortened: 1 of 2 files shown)


sjanssen2:
See discussion in PR #65 . I think we can close PR #65 without merging it. But I'd like to copy the observation about results to here:



@qiyunzhu here are the benchmark result: with the test dataset you already implemented in the unit test:

2 cores, 5 prototypes from 20 elememts:
no-core: 5.4494, filter-while-loop: 5.3082, filter-then-loop: 5.2166

3 cores, 10 prototypes from 100 elements:
no-core: 26.8818, filter-while-loop: 26.7458, filter-then-loop: 26.5841

The result suggests that filter-then-loop is a fast heuristic solution at the cost of reduced sum of distances.

@sjanssen: The remaining question is: is this difference significant?

qiyunzhu:
@sjanssen2 Sounds a cool idea! At my first glance it should work smoothly and swiftly. I am working on something else now. Let me get back to it soon.

mortonjt:
Concerning profilers, I'd take a look at either [line_profiler](https://github.com/rkern/line_profiler) or [cProfile](https://docs.python.org/3/library/profile.html)

qiyunzhu:
Hi guys @sjanssen2 @josenavas @wasade @mortonjt I am quite sick today. Can't focus on the discussion. Excuse me. Thanks for the discussion!

josenavas:
Sorry to hear that! Rest and hope you feel better soon!

qiyunzhu:
@josenavas Thank you for comforting! I am kind-of back to work now.

qiyunzhu:
@sjanssen2 I think your solution is brilliant! Let me test on the master 96K-genome dataset. I am almost there to run the algorithm.

wasade:
Sorry to hear @qiyunzhu! 

I do really recommend running via cProfile. This can be done very easily with ipython magics (eg `%prun`) and will guide focus as to what is expensive.

qiyunzhu:
@sjanssen2 I have got benchmark results on the whole set of 86,200 bacterial and archaeal genomes. They are impressive:

(n prototypes = 10000, n seeds = 4369 NCBI representative genomes)

without seed (pure prototype selection): wall time = 15.2925 sec, sum of distance = 49737208.9099.
with seeds (Stefan's new implementation): 15.2477 sec (almost the same!), 49312938.6542 (only marginally lower)
with seeds (Qiyun's old implementation): 342.8700 sec (much slower), 49312938.6542 (the same)

Then compare with some arbitrarily selected subsets:
one strain per species: 47522719.8175
completely random: 43331905.8102

Therefore, Stefan's algorithm works great in both speed and accuracy, and result is quite stable against arbitrary seeds. @sjanssen2 Cheers!

@wasade Thanks for your recommendation! Let me look into cProfile!

sjanssen2:
I am glad that it works and produces the same results as your PR but faster. Can you do one more test: use my notebook to first filter the seed elements and than run the prototype selection on the remaining matrix without seed. That would than answer our question about the significance of the difference of pre-filter or pre-select.

rob-knight:
This sounds great! Congratulations on the speed improvements. Should take us to a million genomes and beyond, right?

[note, you probably think that’s absurd, but consider how absurd a million 16S sequences sounded in 2007 when we did the Lozupone & Knight paper with a “massive” dataset of 22,000…)

On Mar 14, 2017, at 11:06 AM, Stefan Janssen <notifications@github.com<mailto:notifications@github.com>> wrote:


I am glad that it works and produces the same results as your PR but faster. Can you do one more test: use my notebook to first filter the seed elements and than run the prototype selection on the remaining matrix without seed. That would than answer our question about the significance of the difference of pre-filter or pre-select.

qiyunzhu:
@sjanssen2 I did this test twice and both failed... for unknown reason. I am bond by other things at the moment. Will get back to it soon.

qiyunzhu:
@rob-knight Thanks! Yes this algorithm can be expanded to millions of genomes, since it's linear. Calculating MinHash is linear too, but composing MinHash distance matrix isn't. @tanaes would you like to comment?

PR #69, opened 2017-06-01
Codebase reorganization step 1 - renaming

qiyunzhu:
- Renamed "WGS-HGT" as "Horizomer".
 - Created directory "horizomer" and moved old directories into it.


mortonjt:
Looks good to me

PR #73, opened 2018-01-10
build tree of genome set based on NCBI taxonomy

qiyunzhu:
This script involves quite a few scikit-bio TreeNode operations. The logic is:

1. Read NCBI taxdump into a TreeNode.
2. Shrink the tree so that all tips are among the corresponding TaxIDs of the genome list. Remove redundant single-child internal nodes.
3. There are instances that a genome's TaxID is the parental rank of another (e.g., _E. coli_ vs. _E. coli_ str. D12). In such case, raise its children to be its siblings.
4. translate TaxIDs into genome IDs. In cases where multiple genomes correspond to one TaxID, make them a polytomy.

You may test this script using input files:

- /projects/genome_annotation/20170307/tols/taxonomy/g2tid.txt
- /projects/genome_annotation/20170307/tols/taxonomy/nodes.dmp


horizomer/misc/taxonomy/tree_by_taxdump.py
```diff
@@ -0,0 +1,142 @@
+# this script generates a tree of a set genomes based on their NCBI taxonomy
+# (taxdump) annotations
+# usage: python me.py nodes.dmp genome2taxid.tsv output.nwk
```
mortonjt: Reasonable for a first pass.  But I'd suggest using either click or Argparse, so that a help menu is available.

horizomer/misc/taxonomy/tree_by_taxdump.py
```diff
@@ -0,0 +1,142 @@
...
+root = None
+for tid in taxdump:
+    pid = taxdump[tid]['parent']
+    # at root, the taxid and its parental taxid are both 1
+    if tid == pid:
+        root = tid
+    elif pid in taxdump:
+        taxdump[pid]['children'].add(tid)
+    else:
+        raise ValueError('TaxID %s is not defined.' % pid)
```
mortonjt: Can we separate this out into a separate function?  Once that is done, can we create a unittest for this?

horizomer/misc/taxonomy/tree_by_taxdump.py
```diff
@@ -0,0 +1,142 @@
...
+# create the tree from root
+tree = TreeNode(root)
+
+
+# iteratively attach child nodes to parent node
+def iter_node(node):
+    for cid in taxdump[node.name]['children']:
+        child = TreeNode(cid)
+        node.extend([child])
+        iter_node(child)
```
mortonjt: Can we modularize this, and make taxdump a parameter argument?  Once that is done, can we create a unittest for this?

horizomer/misc/taxonomy/tree_by_taxdump.py
```diff
@@ -0,0 +1,142 @@
...
+# note: multiple genome IDs may correspond to one TaxID
+with open(sys.argv[2], 'r') as f:
+    g2tid = dict(x.split('\t') for x in f.read().splitlines())
+tid2gs = {}  # TaxID to genome ID dictionary
+for g, tid in g2tid.items():
+    if tid not in tid2gs:
+        tid2gs[tid] = set([g])
+    else:
+        tid2gs[tid].add(g)
+tids = set(tid2gs.keys())  # TaxIDs to be retained
```
mortonjt: I'm not exactly sure what this code block is doing, but it would also be nice to see this separated into a function / unittest

horizomer/misc/taxonomy/tree_by_taxdump.py
```diff
@@ -0,0 +1,142 @@
...
+# more tips can be removed.
+while True:
+    tips_to_remove = set()
+    for tip in tree.tips():
+        if tip.name not in tids:
+            tips_to_remove.add(tip)
+    n = len(tips_to_remove)
+    if n == 0:
+        break
+    tree.remove_deleted(lambda x: x in tips_to_remove)
```
mortonjt: Same here, function + unittest would be great!

horizomer/misc/taxonomy/tree_by_taxdump.py
```diff
@@ -0,0 +1,142 @@
...
+nodes_to_remove = []
+for node in tree.traverse():
+    if len(node.children) == 1 and node.name not in tids:
+        nodes_to_remove.append(node)
+for node in nodes_to_remove:
+    if node.is_root():
+        tree = node.children[0]
+    else:
+        node.parent.append(node.children[0])
+        node.parent.remove(node)
```
mortonjt: Same here.  Function + unittest would be awesome.  Btw, what is wrong with scikit-bio's prune function?  
If there is a problem, may warrant us to raise an issue.

horizomer/misc/taxonomy/tree_by_taxdump.py
```diff
@@ -0,0 +1,142 @@
...
+# for internal nodes that are in the dataset, raise their childrens to be
+# siblings of them, so that themselves become tips
+nodes_to_collapse = []
+for node in tree.non_tips():
+    if node.name in tids:
+        nodes_to_collapse.append(node)
+for node in nodes_to_collapse:
+    node.parent.extend(node.children)
+    for child in node.children:
+        node.remove(child)
```
mortonjt: Same here.  Function + unittest

horizomer/misc/taxonomy/tree_by_taxdump.py
```diff
@@ -0,0 +1,142 @@
...
+    if tip.name not in tids:
+        raise ValueError('Error: Failed to remove tip %s.' % tip.name)
+for node in tree.non_tips():
+    if node.name in tids:
+        raise ValueError('Error: Failed to convert %s from an internal node '
+                         'to a tip.' % node.name)
+n = tree.count(tips=True) - len(tids)
+if n > 0:
+    raise ValueError('Error: Failed to retain %d TaxIDs in the dataset as '
+                     'tips of the tree.' % n)
```
mortonjt: Same here.  Function + unittests for each value error

horizomer/misc/taxonomy/tree_by_taxdump.py
```diff
@@ -0,0 +1,142 @@
...
+print('The tree now has %d nodes, including %d tips. The tips and the TaxIDs '
+      'in the dataset match.' % (tree.count(), tree.count(tips=True)))
+
+# translate TaxIDs into genome IDs
+# if multiple genome IDs correspond to one TaxID, make them a polytomy
+tips = [x for x in tree.tips()]
+for tip in tips:
+    tip.parent.extend([TreeNode(x) for x in tid2gs[tip.name]])
+    tip.parent.remove(tip)
```
mortonjt: Same here. Function + unittests please

horizomer/misc/taxonomy/tree_by_taxdump.py
```diff
@@ -0,0 +1,142 @@
...
+print('The TaxIDs at the tips were translated into genome IDs.')
+
+# validate that the tree translation is complete
+for tip in tree.tips():
+    if tip.name not in g2tid:
+        raise ValueError('Error: Failed to translate tip %s.' % tip.name)
+n = tree.count(tips=True) - len(g2tid)
+if n > 0:
+    raise ValueError('Error: Failed to retain %d genome IDs as tips of the '
+                     'tree.' % n)
```
mortonjt: Same here. Function + unittests

mortonjt (changes requested):
This is a lot of complex code here, without proper tests or test input.  Given how complex tree operations are, I strongly recommend putting together some unittests, with predefined inputs.  

This important for many reasons, specifically for sanity checking your code, reorganizing it, and future-proofing it.  I recommend taking a look at the [skbio TreeNode tests](https://github.com/biocore/scikit-bio/blob/master/skbio/tree/_tree.py) or the [Gneiss balance tests](https://github.com/biocore/gneiss/blob/master/gneiss/tests/test_balances.py) to get an idea about how to properly test this object.

horizomer/misc/taxonomy/tree_by_taxdump.py
```diff
@@ -0,0 +1,142 @@
+# this script generates a tree of a set genomes based on their NCBI taxonomy
+# (taxdump) annotations
+# usage: python me.py nodes.dmp genome2taxid.tsv output.nwk
+
+import sys
+from skbio import TreeNode
+
+# read NCBI taxonomy hierarchies (nodes.dmp)
```
tanaes: Another option: there is code in ETE3 to do a lot of this. You can, for instance, pass a list of NCBI taxids and [return the minimal tree](http://etetoolkit.org/docs/latest/tutorial/tutorial_ncbitaxonomy.html#getting-ncbi-species-tree-topology)  encompassing these taxids, eliminating the whole tree-building-pruning portion below. I believe ETE3 also now has a to_skbio method for tranlating to an skbio TreeNode object. 

It should then be simple to add genome IDs on as children to their appropriate taxids.

tanaes: screenshot

tanaes: I believe that reproduces the functionality.

wasade: It would be fantastic to finally get the pycogent code ([entry point](https://github.com/pycogent/pycogent/blob/master/cogent/parse/ncbi_taxonomy.py#L251)) for this ported

horizomer/misc/taxonomy/tree_by_taxdump.py
wasade: suggest just reusing / adapting the existing code the lab produced for this in pycogent

horizomer/misc/taxonomy/tree_by_taxdump.py
wasade: @qiyunzhu, minor but just FYI, `set(tid2gs) == set(tid2gs.keys())`

horizomer/misc/taxonomy/tree_by_taxdump.py
```diff
@@ -0,0 +1,142 @@
...
+    if tid not in tid2gs:
+        tid2gs[tid] = set([g])
+    else:
+        tid2gs[tid].add(g)
+tids = set(tid2gs.keys())  # TaxIDs to be retained
+print('The dataset has %d genomes assigned to %d TaxIDs.'
+      % (len(g2tid), len(tids)))
+
+
+# shrink the taxonomy tree to only contain tips present in the dataset
```
wasade: this sounds like `TreeNode.shear`?

horizomer/misc/taxonomy/tree_by_taxdump.py
```diff
@@ -0,0 +1,142 @@
...
+    n = len(tips_to_remove)
+    if n == 0:
+        break
+    tree.remove_deleted(lambda x: x in tips_to_remove)
+
+print('The taxdump tree was shrinked to %d nodes, including %d tips, all of '
+      'which are in the dataset.' % (tree.count(), tree.count(tips=True)))
+
+# collapse internal nodes with single child and not in the dataset
+# modified from scikit-bio's `prune` function
```
wasade: why not just use `TreeNode.prune`?

qiyunzhu:
Hi @wasade Thanks for reviewing the code!

horizomer/misc/taxonomy/tree_by_taxdump.py
qiyunzhu: Good point!

horizomer/misc/taxonomy/tree_by_taxdump.py
qiyunzhu: I have explored `shear` and `prune` functions, and realized that they don't meet the requirement here.

`prune` deletes all single-child internal nodes. However in this NCBI taxonomy tree, some internal nodes have TaxIDs that are represented in our genome list. For example, _E. coli_  is a parental node of _E. coli str K-12_. In this case we should not delete _E. coli_ even if it's a single-child node.

`shear` calls `prune` when it finishes, thus it cannot be used as well.

horizomer/misc/taxonomy/tree_by_taxdump.py
qiyunzhu: See above.

horizomer/misc/taxonomy/tree_by_taxdump.py
qiyunzhu: @tanaes and @wasade , thanks for pointing me to the existing solutions!

I explored the **ETE** solution. It simply downloads the entire taxdump.tar.gz and builds an sqlite database under my home directory. This behavior is bit too invasive and inflexible. (e.g., I cannot plug my own nodes.dmp, nor can I do unit-tests easily.)

The **pycogent** code Daniel shared appear to be a lightweight and more reasonable solution. Certain codes may need to be updated (e.g., the RankNumbers are hard coded). I think it will be ideal to port it into scikit-bio or other general libraries. For now, let me just do the simplest homebrew way.

qiyunzhu:
I updated the PR by converting this script into production-level codes, as @mortonjt suggested. Now there are three functions with unit-tests.

One new function is `compare_topology`, which tests if two scikit-bio TreeNode objects have identical **topology**, as defined by the topological relationships of node names, regardless of how they are ordered under each node, and branch length.

horizomer/misc/taxonomy/tree_by_taxdump.py
qiyunzhu: @mortonjt Your suggestions are being full-filled. Please check out my new code.

wasade:
Isn't that compare_subsets == 0?

qiyunzhu:
@wasade Thanks for reminding me this! I read the documentation of `compare_subsets` and wasn't sure what it does. But I guess this function as well as `compare_rfd` are probably what I am looking for.

However, after doing some tests I felt that the two functions only concern the relationships among tips, but ignore internal nodes. For example,
```
>>> tree1 = TreeNode.read(['((a,b)e,(c,d)f);'])
>>> tree2 = TreeNode.read(['((a,b)f,(c,d)e);'])
>>> tree1.compare_subsets(tree2)
0.0
```

In the current use case internal nodes have to be considered, though.

qiyunzhu:
Added a standardized test dataset - ref107 (the 107 NCBI reference genomes). It is going to be used extensively by multiple tests.

horizomer/benchmark/parse_output.py
```diff
@@ -122,12 +122,12 @@ def parse_darkhorse(input_f, output_fp, low_lpi=0.0, high_lpi=0.6):
     hgts = []
     # skip header
     next(input_f)
-    for l in input_f:
-        l = l.strip('\r\n').split('\t')
-        best_hit_ids.add(l[3])
-        if low_lpi < float(l[5]) < high_lpi:
-            hgt = '\t'.join((l[0], l[3], l[12], l[13], l[14],
-                             l[6], l[9], l[4]))
+    for line in input_f:
```
sjanssen2, line 125: I personally don't like those magic numbers. From just looking at the code, I figure you could also read this file in via Pandas.read_csv(..., sep='\t') and than select specific columns by their header names or assign header names.

horizomer/benchmark/parse_output.py
```diff
@@ -152,9 +152,9 @@ def parse_hgtector(input_f):
     """
     hgts = []
     for line in input_f:
-        l = line.strip('\r\n').split('\t')
-        if (len(l) == 15) and (l[7] == '1'):
-            hgt = '\t'.join((l[0], l[12], l[13], l[14], l[10], l[11]))
+        x = line.strip('\r\n').split('\t')
```
sjanssen2, line 155: see above

horizomer/utils/tests/test_tree.py
```diff
@@ -0,0 +1,149 @@
...
+        # there isn't any file to remove at the moment
+        # but in the future there will be
+        rmtree(self.working_dir)
+
+    def test_compare_topology(self):
+        """Test comparing topologies of two trees."""
+        # test identical Newick strings
+        tree1 = '(a,b)c;'
+        tree2 = '(a,b)c;'
+        obs = compare_topology(TreeNode.read([tree1]), TreeNode.read([tree2]))
```
sjanssen2, line 47: just out of curiosity: why do you pass tree1 within a list?

horizomer/utils/tree.py
```diff
@@ -0,0 +1,132 @@
...
+        # check if sorted child names are identical
+        children1 = sorted(node1.children, key=lambda x: x.name)
+        children2 = sorted(node2.children, key=lambda x: x.name)
+        taxa1 = [x.name for x in children1]
+        taxa2 = [x.name for x in children2]
+        if taxa1 != taxa2:
+            return False
+
+        # recursively test all children
+        res = []
```
sjanssen2: how about
```return all([compare_topology(child1, child2) for (child1, child2) in zip(children1, children2)])``` instead of lines 50 to 54 ?

File: horizomer/utils/tree.py [hunk 18]
```diff
@@ -0,0 +1,132 @@
...
+                taxonomic rank
+            'name' : str
+                taxon name, empty if names_fp is None
+            'children' : set of str
+                child taxids
+        }
+    """
+    taxdump = {}
+
+    # format of nodes.dmp: taxid | parent taxid | rank | more info...
```
sjanssen2, line 73: I tend to say use Pandas but I remember some odd formatting issues of NCBI taxdumps?!

horizomer/utils/tree.py
```diff
@@ -0,0 +1,132 @@
...
+    Notes
+    -----
+    Flipping child nodes does not change topology.
+    Branch lengths are ignored.
+    """
+    if len(node1.children) != len(node2.children):
+        return False
+    elif len(node1.children) > 0:
+        # check if sorted child names are identical
+        children1 = sorted(node1.children, key=lambda x: x.name)
```
sjanssen2: what happens if a node name is None?

horizomer/utils/tree.py
```diff
@@ -0,0 +1,132 @@
...
+        children2 = sorted(node2.children, key=lambda x: x.name)
+        taxa1 = [x.name for x in children1]
+        taxa2 = [x.name for x in children2]
+        if taxa1 != taxa2:
+            return False
+
+        # recursively test all children
+        res = []
+        for child1, child2 in zip(children1, children2):
+            res.append(compare_topology(child1, child2))
```
mortonjt: uh oh,  looks like there is a bit of recursion here.  This will not play well with larger tree, as the Python stack will need to be continuously adjusted to handle deep recursion.

I'd suggest swapping this out with a [pre-order traversal](http://scikit-bio.org/docs/latest/generated/skbio.tree.TreeNode.preorder.html#skbio.tree.TreeNode.preorder).  Alternatively we can raise an issue to tackle this later.

horizomer/benchmark/parse_output.py
```diff
@@ -241,16 +241,16 @@ def parse_genemark(input_f, genbank_fp):
...
-                                    int(l[3].lstrip('<>')),
-                                    l[1])
-            for (gene, l) in genes.items():
-                if l[0] == start and l[1] == end and l[2] == strand:
+        elif reading and len(x) == 6 and x[5] == '2':
+            (start, end, strand) = (int(x[2].lstrip('<>')),
+                                    int(x[3].lstrip('<>')),
+                                    x[1])
+            for (gene, x) in genes.items():
+                if x[0] == start and x[1] == end and x[2] == strand:
```
mortonjt, line 253: Preferably, there would be tests for the parsers.  We can also defer this in an issue

mortonjt (approved):
Looks sane to me, but there are a couple of minor issues that may need to be pushed to the issue tracker

horizomer/utils/tests/test_tree.py
mortonjt: Because that's how skbio reads in newick strings (here's a [unittest](https://github.com/biocore/scikit-bio/blob/master/skbio/tree/tests/test_tree.py#L56) for example).

qiyunzhu: Right!

horizomer/benchmark/parse_output.py
qiyunzhu: Those are some very old codes (not sure if coded by me). The reason you see those in this PR is because the latest PEP8 complains about the usage of `l`, since one can confuse it with `1`. Therefore, I had to replace all `l`s with `x`s.

horizomer/utils/tree.py
qiyunzhu: Practically one cannot make a Newick tree with a taxon name being empty. One can only do `tree.fine('a').name = ''` or `None` in Python, and when there is `None` in a list, it cannot be sorted. But I think this exception can't happen in this context.

horizomer/benchmark/parse_output.py
qiyunzhu: There are tests for the parsers, see [here](https://github.com/biocore/horizomer/blob/master/horizomer/benchmark/tests/test_parse_output.py). Those `x`s are here because new PEP8 no longer supports `l` being a legit variable name.

horizomer/utils/tree.py
qiyunzhu: In NCBI taxdump, field delimiters are `\t|\t`, and line ends are `\t|`. I explored pandas' `read_table` but couldn't find an easy way to read it directly.

horizomer/utils/tree.py
qiyunzhu: This is a brilliant idea! Strange that I didn't know this Python `all()` function before. Let me upgrade.

horizomer/utils/tree.py
qiyunzhu: This sounds a good idea. However I just tested and realized that preorder traversal doesn't consider the order of child nodes under one parent node. For example, I did:
```
def compare_topology(tree1, tree2):
    for (node1, node2) in zip(tree1.preorder(), tree2.preorder()):
        if node1.name != node2.name:
            return False
        children1 = sorted(node1.children, key=lambda x: x.name)
        children2 = sorted(node2.children, key=lambda x: x.name)
        taxa1 = [x.name for x in children1]
        taxa2 = [x.name for x in children2]
        if taxa1 != taxa2:
            return False
    return True
```
Then I tested it:
```
tree1 = '(a,b)c;'
tree2 = '(b,a)c;'
compare_topology(TreeNode.read([tree1]), TreeNode.read([tree2]))
```
The answer is False. However they actually should be identical, since `a` and `b` are merely flipped.

qiyunzhu:
Hi @sjanssen2 @mortonjt thanks for reviewing! Code updated.

horizomer/benchmark/parse_output.py
sjanssen2: lol, as if x would be that much better. How about "fields" or "columns"? I foresee PEP complaining about single char variables in the future.

horizomer/utils/tests/test_tree.py
sjanssen2: I wasn't aware of this. Cool. I always took the detour via StringIO:
```TreeNode.read(StringIO('(a,b)c;'))```
And thus we see how useful code reviews are :-)

File: horizomer/utils/tree.py (hunk 19 above)
Comment (sjanssen2) [reply horizomer/utils/tree.py]: I disagree: the Greengenes reference tree for SEPP contains a lot of internal nodes without names. Or another example: ```(((a:1,b:2):6,(c:3,d:4):7):8,e:5):9;```
I had a lot of trouble with those unnamed nodes :-/

File: horizomer/utils/tree.py (hunk 18 above)
Comment (sjanssen2) [reply horizomer/utils/tree.py:73]: OK

File: horizomer/utils/tree.py [hunk 22]
```diff
@@ -0,0 +1,130 @@
...
+# ----------------------------------------------------------------------------
+
+#
+# functions relevant to phylogenetic tree operations
+#
+
+from skbio import TreeNode
+
+
+def compare_topology(node1, node2):
```
sjanssen2: This function here poses an interesting algorithmic problem, therefore I had some additional thoughts about it. Not sure if it is worth to explore them in thus PR, maybe better ensure that the current code is doing what is shall do and later trigger a new PR?

At some point you depend on sorting the children nodes alphanumerically with regards to their names. Can it happen, that two children have identical names? If yes, if you recurse further down, how can you ensure that you are looking at the right child? I.e. is your sorting "stable"?

I agree with @mortonjt that recursions are not the strength of python and should be avoided as much as possible. However, due to the re-ordering of child nodes, I don't see a direct way to use pre- or postorder traversals for this use case.

However, if we assume every node has a unique name, we might solve the problem with two traversals per tree:
1) preorder traversal: reorder children by lexicographic sorting
2) preorder traversal: concatenate all node names (delimited with a char not in node names like '#') into one long string
3) compare those strings for tree1 and tree2.

sjanssen2: can't stop thinking about this problem. Correct me if I am wrong, but for topology comparison you are basically only interested in the parent-child relationships, right? If this assumption holds you could do the following:
1) preorder traverse tree1 and store all (parent.name: child.name) relations in one dict
2) preorder traverse tree2 and for every found parent.name: child.name relation, check if it is contained in the dict and keys and values match.
Thus, you avoid explicit sorting and recursion

qiyunzhu: Hi @sjanssen2 thanks for your very insightful thoughts!

Yes it can happen that two children have identical names. Actually for many situations, internal nodes don't have labels (or labels == bootstrap values). In those circumstances, this function no longer works, but scikit-bio's compare_subsets or compare_rfd are better choices. Therefore, the current function assumes that all internal nodes have unique node IDs.

Your idea of concatenated string comparison is interesting! At my first glance it should be much more efficient. Regarding string comparison I think we may design very carefully. For example, to my understanding, for tree `((d,e)b,(f,g)c)a;`, the concatenated string will be `abdecfg`. However, for tree `((d,e,c)b,(g)f)a;` which has a different topology, the string will be the same. Maybe I didn't understand your idea correctly?

Moreover, I guess your last idea regarding dict key - value comparison will work. Actually for the other function I wrote that builds scikit-bio tree from taxdump, I reversed the process you proposed - builds tree based on a parent - children dict.

horizomer/utils/tree.py
qiyunzhu: Please see below my long response to your long comment.

horizomer/utils/tests/test_tree.py
qiyunzhu: 👍

horizomer/benchmark/parse_output.py
qiyunzhu: I feel that `fields` or `columns` are too long are temporary variable names. single-letter names like `x` have been used extensively by people already, such as `[x.name for x in tree.tips()]`. People will complain the PEP8 prohibits this (well people are already complainin about PEP8 itself).

horizomer/utils/tree.py
qiyunzhu: @sjanssen2 I just tested your last idea by this:
```
    dic1 = {node.name: set([x.name for x in node.children])
            for node in node1.non_tips()}
    dic2 = {node.name: set([x.name for x in node.children])
            for node in node2.non_tips()}
    return dic1 == dic2
```
But it didn't generate the correct result. Maybe I didn't do it right...

qiyunzhu: @sjanssen2 okay I think I figured out two solutions, one testing node-parent and the other testing node-children. Both passed unit tests:
1
```
dic1, dic2 = ({node.name: set([child.name for child in node.children])
              for node in tree.traverse() if not node.is_tip()}
              for tree in (tree1, tree2))
return dic1 == dic2
```
2
```
dic1, dic2 = ({node.name: node.parent.name
              for node in tree.traverse() if not node.is_root()}
              for tree in (tree1, tree2))
return dic1 == dic2
```

qiyunzhu:
@sjanssen2 updated the `compare_topology` function based on your inspiration. Thank you!

horizomer/benchmark/parse_output.py
sjanssen2: whatever, this is absolutely not blocking. Ignore this discussion!

sjanssen2:
I am glad that my rough idea worked out. Ready to go in my opinion!

sjanssen2 (approved):

sjanssen2:
looks like I don't have merge rights

PR #74, opened 2018-01-24
added intersecting trees function

qiyunzhu:
A small, simple function that keeps only overlapping taxa of two trees.

### DIFF
--- a/horizomer/utils/tree.py
+++ b/horizomer/utils/tree.py
@@ -15,6 +15,25 @@
 from skbio import TreeNode
 
 
+def has_duplicates(tree):
+    """Test whether there are duplicated taxa (tip names) in a tree.
+
+    Parameters
+    ----------
+    tree : skbio.TreeNode
+        tree for check for duplicates
+
+    Returns
+    -------
+    bool
+        whether there are duplicates
+    """
+    taxa = [tip.name for tip in tree.tips()]
+    if '' in taxa or None in taxa:
+        raise ValueError('Empty taxon name(s) found.')
+    return len(set(taxa)) < len(taxa)
+
+
 def compare_topology(tree1, tree2):
     """Test whether the topologies of two trees with all nodes assigned
     unique IDs are identical.
@@ -59,6 +78,9 @@ def intersect_trees(tree1, tree2):
     tuple of two TreeNodes
         resulting trees containing only overlapping taxa
     """
+    for tree in (tree1, tree2):
+        if has_duplicates(tree):
+            raise ValueError('Either tree has duplicated taxa.')
     taxa1 = set([tip.name for tip in tree1.tips()])
     taxa2 = set([tip.name for tip in tree2.tips()])
     taxa_lap = taxa1.intersection(taxa2)
--- a/horizomer/utils/tree.py
+++ b/horizomer/utils/tree.py
@@ -44,6 +44,31 @@ def compare_topology(tree1, tree2):
     return n2p1 == n2p2
 
 
+def intersect_trees(tree1, tree2):
+    """Shrink two trees to contain only overlapping taxa.
+
+    Parameters
+    ----------
+    tree1 : skbio.TreeNode
+        first tree to intersect
+    tree2 : skbio.TreeNode
+        second tree to intersect
+
+    Returns
+    -------
+    tuple of two TreeNodes
+        resulting trees containing only overlapping taxa
+    """
+    taxa1 = set([tip.name for tip in tree1.tips()])
+    taxa2 = set([tip.name for tip in tree2.tips()])
+    taxa_lap = taxa1.intersection(taxa2)
+    if len(taxa_lap) == 0:
+        raise ValueError('Trees have no overlapping taxa.')
+    tree1_lap = tree1.shear(taxa_lap)
+    tree2_lap = tree2.shear(taxa_lap)
+    return (tree1_lap, tree2_lap)
+
+
 def read_taxdump(nodes_fp, names_fp=None):
     """Read NCBI taxdump.
 
--- a/horizomer/utils/tests/test_tree.py
+++ b/horizomer/utils/tests/test_tree.py
@@ -15,6 +15,7 @@ from os.path import join, dirname, realpath
 from skbio import TreeNode
 
 from horizomer.utils.tree import (compare_topology,
+                                  intersect_trees,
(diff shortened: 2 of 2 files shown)


horizomer/utils/tree.py
```diff
@@ -44,6 +44,31 @@ def compare_topology(tree1, tree2):
     return n2p1 == n2p2
 
 
+def intersect_trees(tree1, tree2):
```
sjanssen2, line 66: high level question? Why do you need to do this? If it is comparing via RF or subsets in skbio, this is done implicitly. You might also want to check skbio's codebase if you cannot "steal" this functionality from there.

horizomer/utils/tree.py
```diff
@@ -44,6 +44,31 @@ def compare_topology(tree1, tree2):
...
+        first tree to intersect
+    tree2 : skbio.TreeNode
+        second tree to intersect
+
+    Returns
+    -------
+    tuple of two TreeNodes
+        resulting trees containing only overlapping taxa
+    """
+    taxa1 = set([tip.name for tip in tree1.tips()])
```
sjanssen2: what if tip.name is None?

horizomer/utils/tree.py
```diff
@@ -44,6 +44,31 @@ def compare_topology(tree1, tree2):
...
+    taxa2 = set([tip.name for tip in tree2.tips()])
+    taxa_lap = taxa1.intersection(taxa2)
+    if len(taxa_lap) == 0:
```
sjanssen2: logical understanding: what if tree1 has one tip with name 'a' and tree2 two tips (but at different sub-trees) with name 'a'?

horizomer/utils/tree.py
```diff
@@ -44,6 +44,31 @@ def compare_topology(tree1, tree2):
...
+        raise ValueError('Trees have no overlapping taxa.')
+    tree1_lap = tree1.shear(taxa_lap)
```
sjanssen2, line 89: shouldn't you also call http://scikit-bio.org/docs/0.5.1/generated/skbio.tree.TreeNode.prune.html#skbio.tree.TreeNode.prune after shearing?

sjanssen2 (commented):
some logical questions

horizomer/utils/tree.py
qiyunzhu: A tip must have a name, unless one sets it to be `None` manually.

horizomer/utils/tree.py
qiyunzhu: I can't find if there is an existing scikit-bio function to do this. Do you know?

horizomer/utils/tree.py
qiyunzhu: scikit-bio's [shear function](https://github.com/biocore/scikit-bio/blob/master/skbio/tree/_tree.py#L470) already does prune in the end.

horizomer/utils/tree.py
sjanssen2: nope:
```
t = TreeNode.read(['(a,);'])
for tip in t.tips():
    print(tip.name)
```

horizomer/utils/tree.py
qiyunzhu: This will break the function. However, I plan to used another function to check if a tree have unique taxon names prior to this function.

qiyunzhu:
@sjanssen2 Thank you! Replies added.

horizomer/utils/tree.py
sjanssen2: I guess it is this one https://github.com/biocore/scikit-bio/blob/0a9a9f013fe445860b741ac949eceac56b2a6fbb/skbio/tree/_tree.py#L2725

horizomer/utils/tree.py
sjanssen2: ups, than I can save some time in some of my functions :-)

horizomer/utils/tree.py
sjanssen2: You might want to add a comment about this in the docstring then

horizomer/utils/tree.py
qiyunzhu: Wow, it looks that my function was trying to replicate the `compare_tip_distances` function. However that function does not return two trees. So I have to write one anyway.

horizomer/utils/tree.py
qiyunzhu: Good experiment!

I think if we test tip name uniqueness explicitly, this `None` check can be omitted.

horizomer/utils/tree.py
qiyunzhu: Let me just add a function to test uniqueness.

horizomer/utils/tree.py
```diff
@@ -44,6 +44,31 @@ def compare_topology(tree1, tree2):
...
+    Returns
+    -------
+    tuple of two TreeNodes
+        resulting trees containing only overlapping taxa
+    """
+    taxa1 = set([tip.name for tip in tree1.tips()])
+    taxa2 = set([tip.name for tip in tree2.tips()])
+    taxa_lap = taxa1.intersection(taxa2)
+    if len(taxa_lap) == 0:
+        raise ValueError('Trees have no overlapping taxa.')
```
tkosciolek, line 88: why do you raise an error here? wouldn't it be more useful to return 0 and give a warning?

tkosciolek (approved):
otherwise looks good

horizomer/utils/tree.py
qiyunzhu: That's because the return structure is a tuple of (TreeNode, TreeNode). User will typically do `x1, y1 = intersec_trees(x0, y0)`. To prevent further errors, better raise an error here.

tkosciolek: ok, fair.
ready to merge?

qiyunzhu: No. Let me add something. @tkosciol

qiyunzhu:
@sjanssen2 @tkosciol I addressed your comments by adding a `has_duplicates` function. Would you like to check again? Thanks!

horizomer/utils/tree.py
qiyunzhu: Addressed this comment.

sjanssen2 (approved):
:+1:

PR #75, opened 2018-01-26
added collapsing functions

qiyunzhu:
Four very useful functions, the latter two of which depend on the former two, and will be important in generating reliable reference trees. They themselves are very simple. But the logics behind them need some consideration.
 - get node support value
 - collapse a node
 - collapse all nodes with branch length < cutoff
 - collapse all nodes with support value < cutoff

### DIFF
--- a/horizomer/utils/tests/test_tree.py
+++ b/horizomer/utils/tests/test_tree.py
@@ -90,114 +90,116 @@ class TreeTests(TestCase):
     def test_has_duplicates(self):
         """Test checking for duplicated taxa."""
         # test tree without duplicates
-        tree = '((a,b),(c,d));'
-        obs = has_duplicates(TreeNode.read([tree]))
+        tree = TreeNode.read(['((a,b),(c,d));'])
+        obs = has_duplicates(tree)
         self.assertFalse(obs)
 
         # test tree with duplicates
-        tree = '((a,a),(c,a));'
-        obs = has_duplicates(TreeNode.read([tree]))
+        tree = TreeNode.read(['((a,a),(c,a));'])
+        obs = has_duplicates(tree)
         self.assertTrue(obs)
 
-        tree = '((1,(2,x)),4,(5,(6,x,8)));'
-        obs = has_duplicates(TreeNode.read([tree]))
+        tree = TreeNode.read(['((1,(2,x)),4,(5,(6,x,8)));'])
+        obs = has_duplicates(tree)
         self.assertTrue(obs)
 
         # test tree with empty taxon names (not a common scenario but can
         # happen)
-        tree = '((1,(2,,)),4,(5,(6,,8)));'
+        tree = TreeNode.read(['((1,(2,,)),4,(5,(6,,8)));'])
         msg = 'Empty taxon name\(s\) found.'
         with self.assertRaisesRegex(ValueError, msg):
-            has_duplicates(TreeNode.read([tree]))
+            has_duplicates(tree)
 
     def test_compare_topology(self):
         """Test comparing topologies of two trees."""
         # test identical Newick strings
-        tree1 = '(a,b)c;'
-        tree2 = '(a,b)c;'
-        obs = compare_topology(TreeNode.read([tree1]), TreeNode.read([tree2]))
+        tree1 = TreeNode.read(['(a,b)c;'])
+        tree2 = TreeNode.read(['(a,b)c;'])
+        obs = compare_topology(tree1, tree2)
         self.assertTrue(obs)
 
         # test identical topologies with different branch lengths
-        tree1 = '(a:1,b:2)c:3;'
-        tree2 = '(a:3,b:2)c:1;'
-        obs = compare_topology(TreeNode.read([tree1]), TreeNode.read([tree2]))
+        tree1 = TreeNode.read(['(a:1,b:2)c:3;'])
+        tree2 = TreeNode.read(['(a:3,b:2)c:1;'])
+        obs = compare_topology(tree1, tree2)
         self.assertTrue(obs)
 
         # test identical topologies with flipped child nodes
-        tree1 = '(a,b)c;'
-        tree2 = '(b,a)c;'
-        obs = compare_topology(TreeNode.read([tree1]), TreeNode.read([tree2]))
+        tree1 = TreeNode.read(['(a,b)c;'])
+        tree2 = TreeNode.read(['(b,a)c;'])
+        obs = compare_topology(tree1, tree2)
         self.assertTrue(obs)
 
-        tree1 = '((4,5)2,(6,7,8)3)1;'
-        tree2 = '((8,7,6)3,(5,4)2)1;'
-        obs = compare_topology(TreeNode.read([tree1]), TreeNode.read([tree2]))
+        tree1 = TreeNode.read(['((4,5)2,(6,7,8)3)1;'])
+        tree2 = TreeNode.read(['((8,7,6)3,(5,4)2)1;'])
+        obs = compare_topology(tree1, tree2)
         self.assertTrue(obs)
 
-        tree1 = '(((9,10)4,(11,12,13)5)2,((14)6,(15,16,17,18)7,(19,20)8)3)1;'
-        tree2 = '(((15,16,17,18)7,(14)6,(20,19)8)3,((12,13,11)5,(10,9)4)2)1;'
-        obs = compare_topology(TreeNode.read([tree1]), TreeNode.read([tree2]))
+        tree1 = TreeNode.read(['(((9,10)4,(11,12,13)5)2,((14)6,(15,16,17,18)7,'
+                               '(19,20)8)3)1;'])
+        tree2 = TreeNode.read(['(((15,16,17,18)7,(14)6,(20,19)8)3,((12,13,11)5'
+                               ',(10,9)4)2)1;'])
+        obs = compare_topology(tree1, tree2)
         self.assertTrue(obs)
 
         # test different topologies
-        tree1 = '(a,b)c;'
-        tree2 = '(a,c)b;'
-        obs = compare_topology(TreeNode.read([tree1]), TreeNode.read([tree2]))
+        tree1 = TreeNode.read(['(a,b)c;'])
+        tree2 = TreeNode.read(['(a,c)b;'])
+        obs = compare_topology(tree1, tree2)
         self.assertFalse(obs)
 
-        tree1 = '((4,5)2,(6,7,8)3)1;'
-        tree2 = '((4,5)3,(6,7,8)2)1;'
-        obs = compare_topology(TreeNode.read([tree1]), TreeNode.read([tree2]))
+        tree1 = TreeNode.read(['((4,5)2,(6,7,8)3)1;'])
+        tree2 = TreeNode.read(['((4,5)3,(6,7,8)2)1;'])
+        obs = compare_topology(tree1, tree2)
         self.assertFalse(obs)
 
-        tree1 = '((4,5)2,(6,7,8)3)1;'
-        tree2 = '(((4,1)8)7,(6,3)2)5;'
-        obs = compare_topology(TreeNode.read([tree1]), TreeNode.read([tree2]))
+        tree1 = TreeNode.read(['((4,5)2,(6,7,8)3)1;'])
+        tree2 = TreeNode.read(['(((4,1)8)7,(6,3)2)5;'])
+        obs = compare_topology(tree1, tree2)
         self.assertFalse(obs)
 
     def test_intersect_trees(self):
         """Test intersecting two trees."""
         # test trees with identical taxa
-        tree1 = '((a,b),(c,d));'
-        tree2 = '(a,(b,c,d));'
-        obs = intersect_trees(TreeNode.read([tree1]), TreeNode.read([tree2]))
-        exp = (TreeNode.read([tree1]), TreeNode.read([tree2]))
+        tree1 = TreeNode.read(['((a,b),(c,d));'])
+        tree2 = TreeNode.read(['(a,(b,c,d));'])
+        obs = intersect_trees(tree1, tree2)
+        exp = (tree1, tree2)
         for i in range(2):
             self.assertEqual(obs[i].compare_subsets(exp[i]), 0.0)
 
         # test trees with partially different taxa
-        tree1 = '((a,b),(c,d));'
-        tree2 = '((a,b),(c,e));'
-        obs = intersect_trees(TreeNode.read([tree1]), TreeNode.read([tree2]))
-        tree1_lap = '((a,b),c);'
-        tree2_lap = '((a,b),e);'
-        exp = (TreeNode.read([tree1_lap]), TreeNode.read([tree2_lap]))
+        tree1 = TreeNode.read(['((a,b),(c,d));'])
+        tree2 = TreeNode.read(['((a,b),(c,e));'])
+        obs = intersect_trees(tree1, tree2)
+        tree1_lap = TreeNode.read(['((a,b),c);'])
+        tree2_lap = TreeNode.read(['((a,b),e);'])
+        exp = (tree1_lap, tree2_lap)
         for i in range(2):
             self.assertEqual(obs[i].compare_subsets(exp[i]), 0.0)
 
-        tree1 = '(((a,b),(c,d)),((e,f,g),h));'
-        tree2 = '(a,((b,x),(d,y,(f,g,h))));'
-        obs = intersect_trees(TreeNode.read([tree1]), TreeNode.read([tree2]))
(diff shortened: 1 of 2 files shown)


horizomer/utils/tree.py
```diff
@@ -15,6 +15,59 @@
...
+    Notes
+    -----
+    A "support value" is defined as the numeric form of a whole node label
+    without ":", or the part preceding the first ":" in the node label.
+    - For examples: "(a,b)1.0", "(a,b)1.0:2.5", and "(a,b)'1.0:species_A'". In
+    these cases the support values are all 1.0.
+    - For examples: "(a,b):1.0" and "(a,b)species_A". In these cases there are
+    no support values.
+    """
+    try:
```
sjanssen2, line 40: that seems to be an ugly hack. Is TreeNode not able to parse support values in the appropriate way? I know it is kind of limited. Maybe you want to parse the tree via ete and than convert the object to TreeNode, which can have additional parameters on each node, like ```node.support = 1.0```

horizomer/utils/tree.py
```diff
@@ -15,6 +15,59 @@
...
+        return float(node.name.split(':')[0])
+    except (ValueError, AttributeError):
+        return None
+
+
+def collapse(node):
+    """Collapse an internal node of a tree.
```
sjanssen2: I don't get what the functions purpose should be from reading the doc string. Can you give an example?

sjanssen2: Am I correct to say that children are moved up to the parent while self get's deleted?
What is the usecase? Do you want to remove internal nodes with just once child?

horizomer/utils/tree.py
```diff
@@ -91,6 +144,66 @@ def intersect_trees(tree1, tree2):
...
 
+def collapse_short_branches(tree, cutoff):
+    """Collapse internal nodes with branch length below cutoff.
+
+    Parameters
+    ----------
+    tree : skbio.TreeNode
+        tree to search for nodes to collapse
+    cutoff : int or float
+        branch length cutoff under which node should be collapsed
```
sjanssen2: node_s_! plural, right?

horizomer/utils/tree.py
```diff
@@ -91,6 +144,66 @@ def intersect_trees(tree1, tree2):
...
+        resulting tree with short branches collapsed
+
+    Notes
+    -------
+    Nodes without branch length are considered as having zero branch length and
+    will be collapsed.
+    """
+    tcopy = tree.copy()
+    nodes_to_collapse = []
+    for node in tcopy.non_tips():
```
sjanssen2: why do you exclude tips? Don't they also have lengths, which might be so small that you need to "merge" them with parent?

horizomer/utils/tree.py
```diff
@@ -91,6 +144,66 @@ def intersect_trees(tree1, tree2):
...
+        if (node.length or 0.0) < cutoff:
+            nodes_to_collapse.append(node)
+    for node in nodes_to_collapse:
```
sjanssen2: Can we encounter recursive situations?

horizomer/utils/tree.py
```diff
@@ -91,6 +144,66 @@ def intersect_trees(tree1, tree2):
...
+        collapse(node)
+    return tcopy
+
+
+def collapse_low_support_nodes(tree, cutoff):
```
sjanssen2: very much reads like a code duplication which you might want to avoid by generalizing both collapse functions?

sjanssen2 (changes requested):
I am in an early stage where I need some enlightenment on the overall logic

qiyunzhu:
@sjanssen2 Thank you for reviewing! Let me try to answer your questions.

horizomer/utils/tree.py
qiyunzhu: The usecase is that one wants to collapse nodes with low support value or short branch length, two signs of unreliable topology. The two other functions are coded for this purpose. For example in this tree:
```
                    /-a
          /c-------|
         |          \-b
---------|
         |          /-d
          \f-------|
                    \-e
```
Assume node `c` has a very low support value. We cannot trust this topology. If the tree is fed to downstream applications like UniFrac, this unreliability may cause problem. So we want to collapse this node, and the topology will become:
```
          /-a
         |
---------|--b
         |
         |          /-d
          \f-------|
                    \-e
```

horizomer/utils/tree.py
qiyunzhu: I think it should be singular in this context? It refers to _a_ node that needs to be collapse because its branch length is below cutoff.

horizomer/utils/tree.py
qiyunzhu: Tips have lengths, however they won't interfere with the topology. If we collapse tips they basically become zero branch length. But the topology remains.

horizomer/utils/tree.py
qiyunzhu: Actually I did recursion in the very beginning. Then I thought a little bit and believed that there is no need for recursion. The first scan will collapse all short branches already. It is not possible that a tree after one round of collapse will have new branches that are below cutoff.

But please correct me if you think otherwise.

horizomer/utils/tree.py
qiyunzhu: That's why I have the `collapse` function in the beginning. But maybe you are saying that these two functions can be further simplified? Such as making another function `collapse_nodes` which takes a lambda function as parameter and takes care of the entire tree?

qiyunzhu: The two functions have slightly different logics. When collapsing short branches, none branch lengths are collapsed too. But when collapsing low support nodes, non support (no label) are kept.

Moreover, I guess that if we remove those two functions, but let user freely combine criteria (e.g., collapsing short branches and low-support nodes and node labels reading "x" so on so forth) there may be logical problems. Although I cannot be very certain of it.

horizomer/utils/tree.py
sjanssen2: I think you should add this beautiful example to the documentation.
And maybe come up with another name like "lift_children" ?

qiyunzhu: Thanks! Sure, I can add it. I think "collapse" is the standard term in phylogenetics.

screen shot 2018-01-26 at 3 22 43 pm

sjanssen2: hm the figtree example shows that "collapse" is merging a sub-tree into a single node or nodelike object. You thus gave a counter example LOL

qiyunzhu: Well... if you just google "phylogenetic tree collapse node" you will get lots of hits. Like this [one](http://treegraph.bioinfweb.info/Help/wiki/Collapsing_nodes_by_support)

horizomer/utils/tree.py
qiyunzhu: As far as I understand, TreeNode cannot handle support values. A support value is but treated as a node label. If we use ETE3 basically we are replacing scikit-bio in all tree operations...

horizomer/utils/tests/test_tree.py
```diff
@@ -39,116 +40,232 @@ def tearDown(self):
...
+
+        # test nodes with support and branch length
+        tree = TreeNode.read(['((a,b)0.85:1.23,(c,d)0.95:4.56);'])
+        node1, node2 = tree.children
+        self.assertEqual(support(node1), 0.85)
+        self.assertEqual(support(node2), 0.95)
+
+        # test nodes with support and extra label (not a common scenario but
+        # can happen)
+        tree = TreeNode.read(['((a,b)\'80:X\',(c,d)\'60:Y\');'])
```
mortonjt, line 58: This will also work in python
```
tree = TreeNode.read(["((a,b)'80:X',(c,d)'60:Y');"])
```
So no need for escape characters (unless you want them).

horizomer/utils/tree.py
mortonjt: yikes!  That is pretty stemming from some the newick parser - it can't handle extra metadata very well.  Probably should raise an issue on this on skbio.

horizomer/utils/tree.py
mortonjt: Agree with @sjanssen2,  collapse can refer to a wide-range of possible operations, including aggregating abundances on tips, merging together nodes into a single node, or creating "super" nodes.

At the very least, I'd suggest inlining that specific example in the documentation.  Renaming the function to `collapse_support` may also help.

horizomer/utils/tree.py
mortonjt: @qiyunzhu definitely ok.  Recursion in Python is not natural, and is recommended to "unroll" the recursion (if I'm understanding this correctly).

horizomer/utils/tree.py
mortonjt: One possibility is to have a function defined as follows

```python
def collapse_by_value(tree, attr, cutoff):
    """ 
    Parameters
    ------------
    tree : skbio.TreeNode
       Tree
    attr : str
       Attribute name
    cutoff : float
       cutoff value
    """
    value = getattr(tree, attr) # can be either support or length
    ...
```
However, this would force you to redecorate all of the support values directly on the trees.  Basically
```
for n in tree.postorder():
   n.support = float(n.name.split(':')[0])
```

Not entirely sure how this fits into the big picture.  But this is one way to reduce code-redundancy.

mortonjt (approved):
Looks like there are some minor things that could be cleaned up.  Overall, it makes sense!

horizomer/utils/tree.py
qiyunzhu: Hi @mortonjt @sjanssen2 how about we do `collapse_node`?

horizomer/utils/tree.py
qiyunzhu: @mortonjt Thanks! In this case I think it is just logically not necessary to recurse or do any advanced tricks.

horizomer/utils/tree.py
qiyunzhu: I think the function suggested by Jamie sounds quite reasonable. Moreover, I think adding `support` as a property to the `TreeNode` class is a good idea. Even though my codes are standalone functions, they can be modified slightly to fit into the framework of scikit-bio. At the current status however, I cannot do `getattr` I think.

Plus, I don't think this will force re-decoration of all nodes. In the collapse functions I wrote, no existing node labels will be modified, including support values.

horizomer/utils/tests/test_tree.py
qiyunzhu: I am trying to uniformly use `'` instead of `"` in all str cases. How does this sound?

horizomer/utils/tree.py
tanaes: Oh man it does seem like handling node-linked metadata is kind of a must for TreeNode. It doesn't seem reasonable to make the TreeNode logic itself limited by the dumpster fire that is the Newick un-format. 

@wasade any thoughts on this?

horizomer/utils/tree.py
tanaes: I'd suggest that the fact that this is stressing our intuitive understanding of the word is indicative of a problem with the terminology, and we should consider that. 

Noodling a bit on this, I think the problem stems from *which* node is being considered here. When we talk about things *collapsing* we usually imagine something larger being condensed into the same thing, only smaller: a star -> a black hole; a building -> a pile of rubble; a cardboard box -> folded cardboard. 

The FigTree usage fits this intution: the *node itself* remains in existence, but the *internal structure* of the node is destroyed. 

The other usage actually destroys the node entirely: it ceases to exist, and the internal structure (the set of descending branches) is removed and separated into no-longer-especially-connected parts. Verbs that more effectively convey this sort of change might be: dissolve, unpack, discombobulate, disaggregate, explode, dissect. 

I'd lean toward **collapse** referring to the maintenance of the node itself, but the dissolution of all internal structure (i.e. all sub-nodes) into one polytomy (or alternative object -- with the recursive node structure this is maybe a less useful operation than in the visual model); and **dissolve** referring to the behavior here, where the immediate sub-branches are grafted onto the parent node. Dissolve might be a bit too strong -- it conveys a *lot* of structural loss, so **disaggregate** or **unpack** might be more apt.

tanaes (commented):
Will continue reviewing later...

horizomer/utils/tree.py
qiyunzhu: Hi @tanaes As far as I understand, the standard Newick format has no place for "node support value" specifically. When you load a tree with bootstraps in FigTree, it asks if the values are "label". In some Newick derivatives (such as BEAST trees), metadata fields are included in `[]`. I think it is ideal though not absolutely essential to add those parsers.

horizomer/utils/tree.py
qiyunzhu: Hi Jon, thanks for the very detailed opinion. I explored a little bit and felt that the terminology is ambiguous. [DendroPy](https://pythonhosted.org/DendroPy/library/treemodel.html?highlight=collapse#dendropy.datamodel.treemodel.Edge.collapse), [Mesquite](https://mesquiteproject.wikispaces.com/Trees) and [TG2](http://treegraph.bioinfweb.info/Help/wiki/Collapsing_nodes_by_support) performs "collapse" in the matter I am doing. Whereas FigTree, Dendroscope and [ETE3](https://github.com/etetoolkit/ete/blob/master/ete3/phylo/phylotree.py#L715) appear to "collapse" an entire clade.

My feeling is that programs that actually manipulate tree topology do "collapse" by deleting a node and raising child nodes, while tree display programs convert a clade into a tip.

My overall suggestion is that we don't invent new terms, but adopt an existing one, i.e., **collapse**, even if its meaning is not crystal clear. What do you think?

sjanssen2: I am not OK with keeping the name "collapse" and I think this controversial discussion is enough indication that the name is a problem. Please change.

horizomer/utils/tree.py
sjanssen2: again, wording might cause the problem. In fact, the function operates on multiple nodes, one of them get's removed but several others get rearranged / extended (length). Arggg, tree operations are not easy. I figure more explicit documentation could help.

horizomer/utils/tree.py
sjanssen2: My gut feeling says that it's risky to exclude this alternative recursion basis. How about an extreme example, where even collapsing the whole tree cannot satisfy the threshold? What would the result be in this case? I would assume it is a single root node with all tips directly attached as children with maximal branch length possible, i.e. also adding up the branch length leading to the tips. Otherwise, path length in the collapsed version would be smaller than in the original topology.

sjanssen2: let's assume a cutoff of 1.0
qiyun

horizomer/utils/tree.py
sjanssen2: I am using attributes extensively. You should be able to do:
`tree.support = 0.123`
or
`print(tree.support)`
with output 0.123
Be aware that tree.support might be `None`

qiyunzhu: Can you show me how to add a property to the TreeNode data type without modifying scikit-bio itself? I wish to learn that.

horizomer/utils/tree.py
qiyunzhu: @sjanssen2 @mortonjt @tanaes I still believe that **collapse** is the most proper term here, especially comparing to English verbs that have not been frequently used by phylogeneticists. However, I do not have strong opinion in terminology, and am willing to make compromise.

Therefore I am proposing to use the function name `collapse_node` here, to describe the action of removing a node and elevating its children, while leaving another function `collapse_clade` for the future, which will convert an entire clade to a single tip. I think these two operations have summarized the two categories of examples I listed above.

horizomer/utils/tree.py
qiyunzhu: Hi @sjanssen2 I tested your case. Result suggests that there is no issue without recursion. See here:
screen shot 2018-02-02 at 1 38 49 pm

horizomer/utils/tree.py
sjanssen2: assume all tree nodes are decorated here with length and/or support (raise an exception otherwise) and check .attribute < cutoff where attribute is a parameter of the function. Thus, you would save code duplication and maybe recomputation of support values for later use

qiyunzhu:
@sjanssen2 I added a function `unpack_by_func` to do both. This is in line with the existing scikit-bio style. For example this `find_by_func` [function](http://scikit-bio.org/docs/0.5.1/generated/skbio.tree.TreeNode.find_by_func.html#skbio.tree.TreeNode.find_by_func).

sjanssen2:
@qiyunzhu looks very good now. Just one thing remains: deletion of the functions that are now superseded with more generic one

qiyunzhu:
@sjanssen2 Thanks! Removed.

sjanssen2 (approved):

PR #76, opened 2018-04-10
Utils

iiiime:
tree.py:
is_nodes_ordered:
replace siblings by children to compare number of child would cause problems
assign prev to None for non_tips() would ignore root
necessary to traverse the tree twice (DFS to assign number of child, BSF for comparison)

test_tree.py:
test_is_nodes_ordered and test_order_nodes updated


qiyunzhu:
Hi @mortonjt , our new volunteer Zijun @nujinuji wrote this code with me. It is to re-order nodes of a tree in increasing or decreasing order. I have reviewed it already (see [here](https://github.com/qiyunzhu/horizomer/pull/2) for my comments). Could you please take a look too? Thank you!

horizomer/utils/tree.py
```diff
@@ -15,6 +15,75 @@
...
+            node.n = sum(x.n for x in node.children)
+    for node in res.postorder():
+        if not node.is_tip():
+            child2n = {x: x.n for x in node.children}
+            node.children = []
+            for child in sorted(child2n, key=child2n.get, reverse=increase):
+                node.append(child)
+    for node in res.postorder():
+        delattr(node, 'n')
+    return res
```
mortonjt, line 47: I'm not sure about the speed, but here is another version (that is fairly efficient).
```
def _cache_ntips(tree):
    for n in tree.postorder(include_self=True):
        if n.is_tip():
            n._n_tips = 1
        else:
            n._n_tips = sum(c._n_tips for c in n.children)
    return tree

def ladderize(tree, ascending=True):
    """
    Sorts tree according to the size of the subtrees.
    Parameters
    ----------
    tree : skbio.TreeNode
       Input tree where leafs correspond to features.
    Returns
    -------
    skbio.TreeNode
       A tree whose tips are sorted according to subtree size.
    Examples
    --------
    >>> from skbio import TreeNode
    >>> tree = TreeNode.read([u'((a,b)c, ((g,h)e,f)d)r;'])
    >>> print(tree.ascii_art())
                        /-a
              /c-------|
             |          \-b
    -r-------|
             |                    /-g
             |          /e-------|
              \d-------|          \-h
                       |
                        \-f
    >>> sorted_tree = ladderize(tree)
    >>> print(sorted_tree.ascii_art())
                        /-a
              /c-------|
             |          \-b
    -r-------|
             |          /-f
              \d-------|
                       |          /-g
                        \e-------|
                                  \-h
    """
    sorted_tree = tree.copy()
    sorted_tree = _cache_ntips(tree)

    for n in sorted_tree.postorder(include_self=True):
        sizes = [k._n_tips for k in n.children]
        idx = np.argsort(sizes)
        if not ascending:
            idx = idx[::-1]
        n.children = [n.children[i] for i in idx]
    return sorted_tree
```

horizomer/utils/tree.py
```diff
@@ -15,6 +15,75 @@
...
+    ----------
+    tree : skbio.TreeNode
+        tree to order
+    increase : bool, optional
+        order nodes in increasing (True) or decreasing (False) order
+
+    Returns
+    -------
+    skbio.TreeNode
+        resulting ordered tree
```
mortonjt, line 31: a doctest on how to use this method would be great.

horizomer/utils/tests/test_tree.py
```diff
@@ -327,6 +328,34 @@ def test_build_taxdump_tree(self):
                              '(19,20)8)3)1;'])
         self.assertTrue(compare_topology(obs, exp))
 
+    def test_order_nodes(self):
+        """Test order nodes"""
+        tree1 = TreeNode.read(['(((a,b),(c,d,i)j),((e,g),h));'])
```
mortonjt, line 333: I would add a more explicit comparison to a tree that is definitely sorted.

For instance, you should try to use this tree and see if this is ordered (it should return true)
```
expected_tree = TreeNode.read(['(((c,d,i)j, (a,b)),((e,g),h));'])
```

mortonjt (approved):
I think this is basically good to go once documentation and tests are addressed.

horizomer/utils/tests/test_tree.py
qiyunzhu: @nujinuji Basically it means that you should provide several trees for each of which you know whether it is ordered for sure, and indicate that in comments.

horizomer/utils/tests/test_tree.py
```diff
@@ -327,6 +328,34 @@ def test_build_taxdump_tree(self):
...
+        self.assertTrue(is_ordered(tree1_decrease, False))
+
+    def test_is_ordered(self):
+        """Test if a tree is ordered"""
+        tree1 = TreeNode.read(['((i,j)a,b)c;'])
+        self.assertTrue(is_ordered(tree1))
+        self.assertTrue(is_ordered(tree1, True))
+        self.assertFalse(is_ordered(tree1, False))
+
+        tree2 = TreeNode.read(['(a, b);'])
```
qiyunzhu: Can remove a space here.

horizomer/utils/tree.py
qiyunzhu: @nujinuji Jamie's implementation was tested working. It is basically very similar to our implementation, except for that he used some tricks that are more efficient (e.g., numpy's `argsort` is more efficient than Python `sorted`). It is for your reference. If you like, you can benchmark the two solutions to get an idea of efficiency.

horizomer/utils/tree.py
qiyunzhu: @nujinuji **doctest** is new to me too. If you have time and interest you can explore how this works.

qiyunzhu:
Hi @mortonjt Zijun updated tests following your request. For docstring how about we take care of it later? Shall we merge?

mortonjt:
Merge!

PR #77, opened 2018-04-13
function cladistic and its unit test added

iiiime:
Function cladistic and its unit test added

### DIFF
--- a/horizomer/utils/tests/test_tree.py
+++ b/horizomer/utils/tests/test_tree.py
@@ -17,7 +17,7 @@ from skbio import TreeNode
 from horizomer.utils.tree import (
     support, unpack, has_duplicates, compare_topology, intersect_trees,
     unpack_by_func, read_taxdump, build_taxdump_tree, order_nodes,
-    is_ordered)
+    is_ordered, cladistic)
 
 
 class TreeTests(TestCase):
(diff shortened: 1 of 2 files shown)


horizomer/utils/tests/test_tree.py
```diff
@@ -361,6 +361,22 @@ def test_is_ordered(self):
         tree5 = TreeNode.read(['((h,(e,g)),((a,b),(c,d,i)j));'])
         self.assertTrue(is_ordered(tree5, False))
 
+    def test_cladistic(self):
+        tree1 = TreeNode.read(['((i,j)a,b)c;'])
```
tkosciolek, line 365: why don't you define trees in setUp, instead? I see you're using the same tree in `test_ordered_nodes`.

tkosciolek (approved):
the code looks good!
just 1 minor comment

horizomer/utils/tests/test_tree.py
qiyunzhu: That's because this whole script has dozens of functions and even more test trees, most of which are distinct. It is just a rare coincidence that these two functions used the same tree.

tkosciolek: ah, ok!

mortonjt (approved):

mortonjt:
Looks good!

PR #78, opened 2018-05-01
added some doctests

qiyunzhu:
- Added doctests for functions `support`, `unpack` and `unpack_by_function`.
 - Modifed `order_nodes`. In the former version, the resulting order is inconsistent when multiple child nodes have the same number of descendant. This was likely because of the usage of dict with TreeNode (which are unorderable) as keys. The fix is to bypass this step and order child nodes directly by the number of descendants. @nujinuji

### DIFF
--- a/horizomer/utils/tree.py
+++ b/horizomer/utils/tree.py
@@ -16,7 +16,8 @@ from skbio import TreeNode
 
 
 def order_nodes(tree, increase=True):
-    """Restore ordering of nodes in one tree based on another tree.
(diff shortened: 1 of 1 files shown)


qiyunzhu:
Hi @nujinuji , could you please take a look to see if my modification is reasonable.
PS: I realized that, to trigger doctests properly, the better way is:
```
python -m doctest horizomer/utils/tree.py 
```

mortonjt (approved):

PR #79, opened 2018-05-11
compare_branch_lengths function and its unit tests added

iiiime:
compare_branch_lengths:
This function determines whether two given trees have same topology and same branch lengths of corresponding nodes. Stack (list) was used to optimize the runtime performance.


qiyunzhu:
@nujinuji and I had several conversations about this function. The goal is to compare two trees to see if they are identical in topology AND in corresponding branch lengths. It cannot differentiate the two situations though. Zijun found the solution with highest performance, with merely *O(n)*, at the cost mentioned above. We think that performance is the top priority here, since we are dealing with large trees. If the user wants, they can perform `compare_rfd` to check topology alone.

horizomer/utils/tree.py
```diff
@@ -13,6 +13,119 @@
...
+    --------
+    >>> from skbio import TreeNode
+    >>> tree = TreeNode.read(["(a:1,b:1)c;"])
+    >>> print(_compare_length(tree.find('a'), tree.find('b')))
+    True
+    >>> print(_compare_length(tree.find('a'), tree.find('c')))
+    False
+
+    """
+    if node1.length is None and node2.length is None:
```
sjanssen2, line 51: can't you just say `return node1.length == node2.length` ?

horizomer/utils/tree.py
```diff
@@ -13,6 +13,119 @@
...
+    >>> from skbio import TreeNode
+    >>> tree1 = TreeNode.read(['((a:1, c:1):2, b:1);'])
+    >>> tree2 = TreeNode.read(['((a:1, c:1):2, b:1);'])
+    >>> print(compare_branch_lengths(tree1, tree2))
+    True
+    >>> tree3 = TreeNode.read(['((a:1, c:1):2, b);'])
+    >>> print(compare_branch_lengths(tree3, tree1))
+    False
+    """
+    tcopy1 = tree1.copy()
```
sjanssen2: copy seems to be a very expensive operation, can you somehow avoid it?

horizomer/utils/tree.py
```diff
@@ -13,6 +13,119 @@
...
+    tcopy2 = tree2.copy()
+    stack = []  # stack to store nodes in tree2
+    count = 0
```
sjanssen2: instead of the count for the for loop you can use `for count, node in enumerate(tcopy1.postorder(include_self=False)):`

horizomer/utils/tree.py
```diff
@@ -13,6 +13,119 @@
...
+
+    for node in tcopy1.postorder(include_self=False):
+        if node.is_tip():
+            try:
+                cur = tcopy2.find(node.name)
```
sjanssen2: why are you searching in the whole tree2, instead of only the corresponding sub-tree of tree2, which you obtain by simultaneously traverse tree1 and tree2, or better their corresponding siblings?

horizomer/utils/tree.py
```diff
@@ -13,6 +13,119 @@
...
+            if node.parent.id is None and cur.parent.id is None:
+                cur.parent.id = node.parent.id = str(count)
+            elif (node.parent.id is not None) ^ (cur.parent.id is not None):
+                return False
+            if cur.parent not in stack:
+                stack.append(cur.parent)
+
+        else:
+            if node.id == stack[-1].id:
+                cur = stack[-1]
```
sjanssen2: why not: `cur = stack.pop()` or `cur = stack.pop(-1)` ?

horizomer/utils/tree.py
```diff
@@ -13,6 +13,119 @@
...
+
+    for node in tcopy1.postorder(include_self=False):
+        if node.is_tip():
+            try:
+                cur = tcopy2.find(node.name)
+            except MissingNodeError:
+                return False
+            if _compare_length(node, cur) is False:
+                return False
+            if node.parent.id is None and cur.parent.id is None:
```
sjanssen2: the next few lines seem to be a code duplication between the cases of being a tip or an internal node. Is it possible to move them up and thus merge?

sjanssen2 (changes requested):
I honestly don't understand the logic of this algorithm. To me, it seems to be too complicated. Here is what I think is the problem:
"Given two _unordered_ trees, test if both follow the same topology AND the same branch length (ignoring node names)". 
The problem is, that you cannot easily determine which of the siblings in the other node correspond to the one child of the current node you are dealing with right now. Looks like you rely on node names here (which contradicts your doc string, since you are only speaking about topology and branch length), which can be risky.

Thus, either you need to improve documentation to reflect the actual problem, or think hard about edge cases!

horizomer/utils/tree.py
iiiime: In case that tree1 and tree2 have different node orderings but same topology, simultaneous traversal of tree1 and tree2 can be hardly achieved. For example:
tree1 = TreeNode.read(['((a:1,(b:1,c:1)d:1)e:1,f:1)g:1;'])
tree2 = TreeNode.read(['(f:1,((b:1,c:1)d:1,a:1)e:1)g:1;'])

While traversing tree1, the location of the corresponding node in tree2 is unknown, so find() is used here. find() creates a tip cache dictionary for the first time of method call, so it is not an expensive operation to search in whole tree 2.

horizomer/utils/tree.py
iiiime: `isclose()` is used here to handle float.

horizomer/utils/tree.py
qiyunzhu: I think this is a good suggestion, and is doable. The only modification we made to the trees is the `id` attribute. I noticed that in the (scikit-bio source code)[(https://github.com/biocore/scikit-bio/blob/master/skbio/tree/_tree.py], many functions call `assign_ids` in the beginning, but don't clean them up in the end. So I think this behavior is tolerable.

horizomer/utils/tree.py
sjanssen2: I agree that simultaneous traversal is not trivial. In fact, you are entering the realm of pairwise tree alignment if you try to solve this problem, which is at least O(n*m) for ordered trees with fixed maximum width and O(n^4) in more relaxed instances.
I wonder if you can rely on node names as anchor points to short-cut the tree alignment problem. If that is the case, you definitively need to elaborate on this major assumption in the docstring!

qiyunzhu:
@sjanssen2 Zijun updated the PR to address your comments. Would you like to take another look? Thank you!

qiyunzhu:
Hi @sjanssen2 I wonder if you have had a chance to take a final look at this PR. I think all your requests have been addressed. If you happen to be inconvenient recently, perhaps I will just merge the PR in order to proceed? Thanks you!

PR #80, opened 2018-05-25
walk_copy, root_above and their unit tests added

iiiime:


horizomer/utils/tests/test_tree.py
```diff
@@ -457,6 +458,128 @@ def test_compare_branch_lengths(self):
         tree11 = TreeNode.read(['((a:1,(x:1,c:1)d:1)e:1,f:1)g:1;'])
         self.assertFalse(compare_branch_lengths(tree11, tree1))
 
+    def test_walk_copy(self):
+        tree1 = TreeNode.read(['(((a:1.0,b:0.8)c:2.4,(d:0.8,e:0.6)f:1.2)g:0.4,'
+                               '(h:0.5,i:0.7)j:1.8)k;'])
+        assign_supports(tree1)
+
+        # test pos = root
+        msg = 'Cannot walk from root of an rooted tree.'
```
qiyunzhu: `an` => `a`.

horizomer/utils/tests/test_tree.py
```diff
@@ -457,6 +458,128 @@ def test_compare_branch_lengths(self):
...
+            walk_copy(tree1.find('f'), tree1.find('k'))
+
+        # test pos = basal
+        with self.assertRaisesRegex(ValueError, msg):
+            walk_copy(tree1.find('g'), tree1.find('a'))
+        with self.assertRaisesRegex(ValueError, msg):
+            walk_copy(tree1.find('g'), tree1.find('k'))
+
+        # pos = derived, move = up
+        res = TreeNode.read(['(a:1.0,((d:0.8,e:0.6)f:1.2,'
```
qiyunzhu: Usually people like to use `obs` (observation) and `exp` (expectation) in a unit test. For example:
```
exp = TreeNode.read(['...'])
obs = walk_copy(x, y)
for n1, n2 in zip(obs.postorder(), exp.postorder()):
    self.assertTrue(n1.name, n2.name)
```

horizomer/utils/tests/test_tree.py
```diff
@@ -457,6 +458,128 @@ def test_compare_branch_lengths(self):
...
+                             '(h:0.5,i:0.7)j:2.2)g:2.4)c:0.8;'])
+        output = walk_copy(tree1.find('c'), tree1.find('a'))
+        for n1, n2 in zip(res.postorder(), output.postorder()):
```
qiyunzhu: I think this comparison can be singled out as an independent function. So that we don't have to duplicate code.

horizomer/utils/tests/test_tree.py
```diff
@@ -457,6 +458,128 @@ def test_compare_branch_lengths(self):
...
+        tree3 = TreeNode.read(['(((a:0.4,b:0.3)e:0.1,(c:0.4,'
+                               'd:0.1)f:0.2)g:0.6)h:0.2;'])
+        assign_supports(tree3)
+
+        tree3_ae = root_above(tree3.find('a'))
+        res = TreeNode.read(['(a:0.2,(b:0.3,((c:0.4,d:0.1)f:0.2,'
+                             'h:0.6)g:0.1)e:0.2);'])
+        for n1, n2 in zip(res.postorder(), tree3_ae.postorder()):
+            self.assertEqual(n1.name, n2.name)
+            self.assertEqual(n1.length, n2.length)
```
qiyunzhu: I don't see any instance that the `support` attribute is being tested. I think we need it.

horizomer/utils/tree.py
```diff
@@ -17,6 +17,253 @@
 from skbio.tree import MissingNodeError
 
 
+def _extract_support(node):
+    """Extract the support value from a node label, if available.
+    Returns
```
qiyunzhu: Missing `Parameters`

horizomer/utils/tree.py
```diff
@@ -17,6 +17,253 @@
...
+    if node.support is not None:  # prevents support of NoneType
+        lblst.append(str(node.support))
+    if node.name:  # prevents name of NoneType
+        lblst.append(node.name)
+    return ':'.join(lblst)
+
+
+def assign_supports(tree):
+    """Extract support values from internal node labels of a tree.
+    Notes
```
qiyunzhu: We need some blank lines here. When you copy something directly from GitHub, the blank lines are gone.

qiyunzhu (commented):
Great job! This is an enormous work. I have a few minor comments.

horizomer/utils/tests/test_tree.py
iiiime: Is it better to add a comparison method in tree.py or test_tree.py?

qiyunzhu: @nujinuji Should be in `tree.py`.

horizomer/utils/tree.py
```diff
@@ -264,6 +279,30 @@ def root_above(node, name=None):
...
+        second tree to compare
+
+    Returns
+    -------
+    bool
+        `True` if name, length, and support of each corresponding node are same
+        `False` otherwise
+    """
+    for n1, n2 in zip(tree1.postorder(), tree2.postorder()):
+        if (n1.name != n2.name or n1.length != n2.length or
```
qiyunzhu: One issue: A TreeNode object by default does not have the `support` attribute. So we need to use `hasattr` to test this before comparing.

qiyunzhu (commented):
One more comment. Everything else is great!

horizomer/utils/tree.py
qiyunzhu: Or maybe use [`getattr`](https://docs.python.org/3/library/functions.html#getattr) if you want to make things simple.

qiyunzhu (commented):
Looks good! However, we also need to have some unit tests to cover those situations. For example, test some trees without support, without branch length, without internal node names, or with them one being `None` and the other being `0.0`, etc. You can just modify your existing code to allow some of these scenarios.

qiyunzhu (approved):
Great job! Let's call it a finish! @nujinuji

